PR #14, opened 2021-03-24
bump kube-prometheus-stack to 14.x for hivemq-operator chart

patrickjahns:
# Motivation
Keep the dependencies in sync between swarm and operator chart 

Probably we should bump the Chart version here - not sure if we use here minor or patch bumps - suggestions?


MagiErber:
@cla-bot check

PR #17, opened 2021-04-26
Add support to select namespaces in HiveMQ Swarm dashboard

dictvm:
Currently, multiple HiveMQ Swarm instances spread over multiple
namespaces cause a) the dashboard to remain empty and b) a templating
error to be displayed.

This modified dashboard adds support for selecting namespaces so that
the graphs are rendered even if multiple Swarm installations exist on
the same cluster but spread over namespaces.


dictvm:
@cla-bot check

PR #25, opened 2021-06-23
Run HiveMQ & operator as non-root user by default

sbaier1:
still a WIP, dependent images still pending release/update

### DIFF
--- a/charts/hivemq-operator/operator-tmpls/cluster-stateful-set.yaml
+++ b/charts/hivemq-operator/operator-tmpls/cluster-stateful-set.yaml
@@ -16,8 +16,14 @@ metadata:
     name: "{{ spec.name }}"
     uid: "{{ spec.metadata.uid }}"
 spec:
-  serviceName: hivemq-{{ spec.name }}-cluster
   replicas: {{ spec.nodeCount }}
+  serviceName: hivemq-{{ spec.name }}-cluster
+  podManagementPolicy: Parallel
+  updateStrategy:
+    type: OnDelete
+  # TODO get this from the CR
+  volumeClaimTemplates:
+    {{ util:nindent(4, util:toYaml(spec.volumeClaimTemplates)) }}
   selector:
     matchLabels:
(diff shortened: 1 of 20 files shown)


charts/hivemq-operator/operator-tmpls/cluster-stateful-set.yaml
```diff
@@ -16,8 +16,14 @@ metadata:
     name: "{{ spec.name }}"
     uid: "{{ spec.metadata.uid }}"
 spec:
-  serviceName: hivemq-{{ spec.name }}-cluster
   replicas: {{ spec.nodeCount }}
+  serviceName: hivemq-{{ spec.name }}-cluster
+  podManagementPolicy: Parallel
+  updateStrategy:
+    type: OnDelete
+  # TODO get this from the CR
```
schaebo, line 24: TODO still valid ?

schaebo (approved):

charts/hivemq-operator/operator-tmpls/cluster-stateful-set.yaml
patrickjahns: The ToDo is also still in the upstream operator template changes - needs to be removed there when we sync the final template manifestes with the helm chart

charts/hivemq-operator/templates/global/hivemq-psp.yaml
```diff
@@ -31,25 +29,25 @@ spec:
   hostPID: false
   runAsUser:
     # Permits the container to run with root privileges as well.
-    rule: 'RunAsAny'
+    rule: 'MustRunAsNonRoot'
```
patrickjahns, line 32: The comment above the line would need adjustment, as we are now explicitly forbid to run as root

patrickjahns:
@sbaier1 
Regarding the admission hooks - I believe we need to adjust the `RunAs` configuration as well ( see: https://github.com/hivemq/helm-charts/pull/25/files )

PR #41, opened 2021-10-28
Add missing claim templates field in CR

sbaier1:
the custom resource template is missing the volumeClaimTemplates field, which is necessary for statefulset resources

### DIFF
--- a/charts/hivemq-operator/templates/hivemq-operator/custom-resource.yaml
+++ b/charts/hivemq-operator/templates/hivemq-operator/custom-resource.yaml
@@ -92,6 +92,8 @@ spec:
   {{- end }}
   initBusyboxImage: {{ .Values.hivemq.initBusyboxImage }}
   initDnsWaitImage: {{ .Values.hivemq.initDnsWaitImage }}
-  volumeClaimTemplates:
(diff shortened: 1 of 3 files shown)


charts/hivemq-operator/templates/hivemq-operator/custom-resource.yaml
```diff
@@ -92,4 +92,6 @@ spec:
   {{- end }}
   initBusyboxImage: {{ .Values.hivemq.initBusyboxImage }}
   initDnsWaitImage: {{ .Values.hivemq.initDnsWaitImage }}
+  volumeClaimTemplates:
+  {{ .Values.hivemq.volumeClaimTemplates }}
```
mhofsche, line 96: Should this be: 
{{- if .Values.hivemq.volumeClaimTemplates }}
volumeClaimTemplates: {{ toYaml .Values.hivemq.volumeClaimTemplates | nindent 4 }}
{{- end }}

sbaier1: could do that, but also `.Values.hivemq.volumeClaimTemplates` has a default value of empty array, so IMO it's not necessary

PR #51, opened 2022-03-25
Allow to pass additional environment variables 

patrickjahns:
# Motivation

We want to pass custom env variables to the operator when deploying via the helm chart

# Descriptions

Adds a variable to the helm-chart operator section and renders respective keys into the deployment

### DIFF
--- a/charts/hivemq-operator/values.yaml
+++ b/charts/hivemq-operator/values.yaml
@@ -58,6 +58,10 @@ operator:
       tolerations: []
   # Set this string to a name of a externally managed ConfigMap containing the templates for the operator if you want to customize them.
   templateConfigMap: ""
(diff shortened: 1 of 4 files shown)


charts/hivemq-operator/values.yaml
```diff
@@ -46,6 +46,10 @@ operator:
       tolerations: []
   # Set this string to a name of a externally managed ConfigMap containing the templates for the operator if you want to customize them.
   templateConfigMap: ""
+  ## Extra environment variables that will be pass to the operator in the form of
```
hurtadosanti, line 49: Can you provide me please a couple of examples, why normal environment variables are not enough?

hurtadosanti (changes requested):
Please update the Changelog

hurtadosanti (approved):

PR #57, opened 2022-05-16
feat: cross namespace mode

sbaier1:
rebase and merge after released.

adds operator cross-namespace support as a field to the helm chart values


hurtadosanti (approved):
We should not add more features without proper testing, I approve this as is needed by the cloud team

Issue #66, opened 2022-07-25
Operator gets killed because it runs out of memory


florianluediger:
When following [these instructions](https://www.hivemq.com/docs/operator/4.8/kubernetes-operator/operator-intro.html), I have noticed that the default settings for the operator do not provide enough memory to run it.

After running `helm upgrade --install hivemq hivemq/hivemq-operator`, the operator pod will be created but when monitoring with K9s, you can see that it crashes pretty soon due to an OOM error.

The pod will restart pretty quickly, however it does not seem to function properly after that.

I suggest increasing the default value for the operator memory limit.

hurtadosanti:
Hi @florianluediger I will review this, we have not seen this behavior yet,  can you provide more context for the error, please?

florianluediger:
Sure, just let me know what you need to know. Here is some more general information:

I have started minikube with 8192M of memory and 4 CPUs. 
Then I have executed `helm upgrade --install hivemq hivemq/hivemq-operator` with no special values provided.

Here is the log of the first start of the hivemq operator pod (grabbed from k9s):
```
Picked up JAVA_TOOL_OPTIONS: -XX:+UnlockExperimentalVMOptions -XX:InitialRAMPercentage=30 -XX:MaxRAMPercentage=80 -XX:MinRAMPercentage=30                                                                       │
09:40:00.261 [main] INFO  com.hivemq.Application - Preparing SSL files                                                                                                                                          │
09:40:08.758 [main] INFO  com.hivemq.Application - Wrote converted key store to /tmp/store.p12                                                                                                                  │
 __  __ _                                  _                                                                                                                                                                    │
|  \/  (_) ___ _ __ ___  _ __   __ _ _   _| |_                                                                                                                                                                  │
| |\/| | |/ __| '__/ _ \| '_ \ / _` | | | | __|                                                                                                                                                                 │
| |  | | | (__| | | (_) | | | | (_| | |_| | |_                                                                                                                                                                  │
|_|  |_|_|\___|_|  \___/|_| |_|\__,_|\__,_|\__|                                                                                                                                                                 │
  Micronaut (v2.4.2)                                                                                                                                                                                            │
                                                                                                                                                                                                                │
Stream closed EOF for default/hivemq-hivemq-operator-operator-6fdb648d47-pbcbb (operator)
```

The log of the operator pod after restarting the pod is empty. It seems like after the restart, the micronaut service does not start anymore.

[...]

PR #67, opened 2022-07-25
Increase operator memory limit to ensure stable operation

florianluediger:
This PR fixes the issue mentioned in #66

### DIFF
--- a/charts/hivemq-operator/values.yaml
+++ b/charts/hivemq-operator/values.yaml
@@ -65,10 +65,10 @@ operator:
   resources:
     limits:
       cpu: 800m
-      memory: 512M
+      memory: 1024M
     requests:
       cpu: 200m
-      memory: 512M
+      memory: 1024M
 monitoring:
   # Deploy ServiceMonitor for Prometheus Operator and enable Prometheus
   enabled: false


florianluediger:
I have already sent the E-Mail with the signed code of conduct to the address mentioned in the link. Will this be updated automatically as soon as the mail is processed or do I have to do something else?

hurtadosanti:
Hi @florianluediger you don't need to modify the `values.yaml` template to achieve this. What is recommended is to use a custom file as described in the documentation https://www.hivemq.com/docs/operator/4.8/kubernetes-operator/deploying.html#deploy-operator  with a command similar to this:
`helm upgrade --install -f myCustomValues.yaml hivemq hivemq/hivemq-operator`

hurtadosanti:
Hello, Florian. Thank you very much for your contribution. Unfortunately, your change is for a very specific use case, so we will reject it for the time being. We suggest overriding the values using the command from the [official documentation](https://www.hivemq.com/docs/operator/4.8/kubernetes-operator/deploying.html#helm-chart)  
  
`helm upgrade --install -f myCustomValues.yaml hivemq hivemq/hivemq-operator`  
  
I will also recommend using smaller cluster and node requirements or provide more memory to the VM 
  
``` 
hivemq:  
  nodeCount: "4" 
  cpu: "4" 
  memory: "4Gi" 
``` 
  
Best wishes,

PR #77, opened 2022-08-19
Increase size of limits and reduce the percentage required

hurtadosanti:


sbaier1 (approved):

sbaier1:
can we also add a parameter to customize it to values.yaml?

PR #83, opened 2022-09-21
chore: Add new enterprise extensions

hurtadosanti:
- Add new enterprise extensions into the values as they are preinstalled
  - hivemq-google-cloud-pubsub-extension
  - hivemq-distributed-tracing-extension
- Bump operator version with multi-namespace fix

### DIFF
--- a/charts/hivemq-operator/Chart.yaml
+++ b/charts/hivemq-operator/Chart.yaml
@@ -10,7 +10,7 @@ keywords:
   - messaging
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.11.7
+version: 0.12.0
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 appVersion: 4.8.4
--- a/charts/hivemq-operator/Chart.yaml
+++ b/charts/hivemq-operator/Chart.yaml
@@ -10,7 +10,7 @@ keywords:
   - messaging
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.11.6
+version: 0.11.7
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 appVersion: 4.8.4
--- a/charts/hivemq-operator/Chart.yaml
+++ b/charts/hivemq-operator/Chart.yaml
@@ -10,7 +10,7 @@ keywords:
   - messaging
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.12.0
+version: 0.11.6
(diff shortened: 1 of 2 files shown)


charts/hivemq-operator/Chart.yaml
```diff
@@ -10,7 +10,7 @@ keywords:
   - messaging
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.11.6
+version: 0.11.7
```
YannickWeber, line 13: Shouldn't this be 0.12.0 then because it adds new functionality?

hurtadosanti: good point, thanks

charts/hivemq-operator/Chart.yaml
```diff
@@ -10,7 +10,7 @@ keywords:
   - messaging
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.11.6
+version: 0.12.0
```
mhofsche, line 13: Is the version numbering change coordinated with Simon?

charts/hivemq-operator/values.yaml
```diff
@@ -187,6 +187,12 @@ hivemq:
 
         [[ ! -f drivers/postgres-jdbc.jar ]] &&
         curl -L https://jdbc.postgresql.org/download/postgresql-42.2.14.jar --output drivers/jdbc/postgres.jar
+    - name: hivemq-google-cloud-pubsub-extension
```
mhofsche, line 190: Can we move the pub/sub extension to just after the kafka extension? That way we can group available enterprise extensions by topic. Once a kinesis becomes available it would move to after pub/sub.

mhofsche (approved):
minor comments

charts/hivemq-operator/Chart.yaml
sbaier1: hm, i'm not sure what you mean by this? so far the versioning "convention" (which isn't documented anywhere AFAIK though) was to bump the minor version when templates change. a quick diff shows though that templates have already changed within 0.11.x so 🤷‍♂️  not sure what to do

i think we can stay within 0.11.x though because no template is changing within this version IMO.

hurtadosanti: Regarding semantic versioning https://semver.org the minor is done when there are not breaking changes but functionality is added, this change is for me on the greyline, yes is only a change on the templates but add functionally, indirectly. Now the questions is then, change on the values does not add functionality but templates do? That seems strange, in either case seems highly subjective.

hurtadosanti: FYI: https://github.com/hivemq/helm-charts/pull/83#discussion_r976096649

PR #106, opened 2023-07-17
Remove initialization configuration and add tests

hurtadosanti:
https://hivemq.kanbanize.com/ctrl_board/22/cards/15585


src/integrationTest/java/com/hivemq/helmcharts/EseExtensionInstallationIT.java
```diff
@@ -0,0 +1,66 @@
...
+
+        final var client = container.getKubernetesClient();
+
+        final URL resource = getClass().getResource("/ese-config.xml");
+
+        final var configMapData = Files.readString(Path.of(Objects.requireNonNull(resource).toURI()));
+        final var eseConfigMap = new ConfigMapBuilder().withNewMetadata()
+                .withName("ese-extension-config")
+                .endMetadata()
+                .withData(Map.of("enterprise-security-extension.xml", configMapData))
```
Donnerbart, line 47:
```suggestion
                .withData(Map.of("hivemq-enterprise-security-extension.xml", configMapData))
```

charts/hivemq-operator/values.yaml
```diff
@@ -187,11 +187,7 @@ hivemq:
       enabled: false
       # Note that this is just an example initialization routine. Make sure this points to the current JDBC version you require for your configuration.
       initialization: |
-        # A little hack because k8s configMaps can't handle sub-directories
```
Donnerbart, line 190: All released HiveMQ versions still support the old config location, and the older versions (4.9 etc.) are _only_ supporting the old config location.

So I think this is the correct way to handle this.

charts/hivemq-operator/values.yaml
```diff
@@ -187,11 +187,7 @@ hivemq:
...
-        [[ -e /conf-override/extensions/hivemq-enterprise-security-extension/enterprise-security-extension.xml ]] &&
-        rm -f $(pwd)/conf/enterprise-security-extension.xml &&
-        cp -s /conf-override/extensions/hivemq-enterprise-security-extension/enterprise-security-extension.xml $(pwd)/conf/enterprise-security-extension.xml
-
+        # K8s configMaps can't handle sub-directories
```
Donnerbart, line 190:
```suggestion
        # Download JDBC driver for PostgreSQL
```

Donnerbart (approved):

charts/hivemq-operator/values.yaml
```diff
@@ -425,7 +421,7 @@ hivemq:
                 <enabled>${HIVEMQ_ALLOW_EMPTY_CLIENT_ID}</enabled>
             </allow-empty-client-id>
 
-            <!-- Configures validation for UTF-8 PUBLISH payloads -->
+            <!-- Configures validation for UTF-8 PUBLISHES payloads -->
```
mhofsche, line 424: let's keep "PUBLISH" here.

charts/hivemq-operator/values.yaml
mhofsche: Please add the answer to this comment here, for future reference

src/integrationTest/resources/values/ese-extension-values.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+operator:
+  image: hivemq/hivemq-operator:snapshot
+  imagePullPolicy: Never
+  logLevel: DEBUG
+  admissionWebhooks:
+    enabled: false
+
+hivemq:
+  image: hivemq/hivemq4-test:k8s-snapshot
```
mhofsche, line 13: what version of hivemq is tested here? Is it the latest? 
Can we add a test with the 4.9.latest image as well?

hurtadosanti: We will need a new card, this only tests locally (so the image is from the master branch) so we can test on the mac and from development

charts/hivemq-operator/values.yaml
hurtadosanti: Yes this was added to avoid the warnings, however it increases complexity to understand which config is being used.

PR #117, opened 2023-10-10
Upgrade to Gradle 8

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/41/cards/17200/details/


hurtadosanti (approved):

build.gradle.kts
```diff
@@ -4,6 +4,12 @@ plugins {
 
 group = "com.hivemq.helmcharts"
 
+java {
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(11))
+    }
+}
+
```
SgtSilvio, line 12: was this now intended to be added?

SgtSilvio (approved):

build.gradle.kts
afalhambra-hivemq: @Donnerbart could you please confirm? I can see it was included in your last commit here.
Though from my point of view, it should be fine and good to have as well

Donnerbart: I've added it here, since we had to remove this setting from the included `hivemq-operator` (because of the Micronaut 3.x Gradle plugin). We have a bunch of Java test classes in this project, I thought it's good to pin them to Java 11.

PR #195, opened 2024-03-13
Add support to configure secure Control Center listeners

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/19797/details/


charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -137,6 +137,22 @@ Returns:
 {{- $typeExists }}
 {{- end -}}
 
+{{/*
+Validate there is no duplicated `containerPort` defined for the services.
```
Donnerbart, line 141: This is really great!

Donnerbart: I think this would also be very helpful for all the other ports we have (partly hard-coded, partly configurable), e.g. cluster transport, Prometheus port, Operator REST API, Health API, ...

Otherwise customers could e.g. try to create an MQTT listener on the cluster transport port or other invalid combinations.

charts/hivemq-platform/tests/hivemq_services_test.yaml
```diff
@@ -159,7 +159,362 @@ tests:
...
+          exposed: true
+          containerPort: 1883
+        - type: control-center
+          exposed: true
+          containerPort: 1883
+    asserts:
+      - failedTemplate:
+          errorMessage: Container port 1883 in service `control-center` already exists
+
+  - it: with duplicated containerPorts in disabled services, then works
```
Donnerbart, line 494: Nice detail!

Donnerbart (approved):
This is a huge improvement!

mhofsche (approved):

charts/hivemq-platform/templates/_helpers.tpl
afalhambra-hivemq: Great suggestion!
Just added an additional validation function to make sure this is also covered. Can you please give a look? Thanks

Donnerbart: I just gave it a quick local spin:
```yaml
  # REST API service configuration
  - type: rest-api
    exposed: true
    containerPort: 7000
```
```shell
donnerbart@firefly:~$ ./install-platform.sh -f platform-values.yaml -l
Using values file: platform-values.yaml
Using local platform chart
Helm command: helm upgrade test-platform hivemq-platform  --install --atomic -n default -f platform-values.yaml  

Release "test-platform" does not exist. Installing it now.
Error: execution error at (hivemq-platform/templates/hivemq-custom-resource.yml:163:8): Container port 7000 in service `rest-api` already exists as part of one of the predefined ports (7979, 8889, 9399, 7000
```
This is amazing!

PR #200, opened 2024-03-19
Remove `isNotEmpty` tag from unit test and replace with `exists`

afalhambra-hivemq:


Donnerbart (approved):
Thanks for taking care of this!

PR #225, opened 2024-04-26
Fix deprecated usage of gradle/gradle-build-action

Donnerbart:
Resolves the following deprecation warnings:
* https://github.com/gradle/actions/blob/main/docs/deprecation-upgrade-guide.md#the-action-gradlegradle-build-action-has-been-replaced-by-gradleactionssetup-gradle
* https://github.com/gradle/actions/blob/main/docs/deprecation-upgrade-guide.md#using-the-action-to-execute-gradle-via-the-arguments-parameter-is-deprecated

### DIFF
--- a/.github/workflows/hivemq-operator-integration-test.yml
+++ b/.github/workflows/hivemq-operator-integration-test.yml
@@ -50,20 +50,19 @@ jobs:
         with:
           java-version: '11'
           distribution: 'temurin'
-      - name: Run HiveMQ Operator (legacy) integration tests
-        uses: gradle/gradle-build-action@e2097ccd7e8ed48671dc068ac4efa86d25745b39 # v3
-        env:
-          TEST_PLAN: ${{ matrix.test-plan }}
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@db19848a5fa7950289d3668fb053140cf3028d43 # v3
         with:
-          cache-disabled: true
           gradle-home-cache-includes: |
             caches
             notifications
             jdks
           gradle-home-cache-cleanup: true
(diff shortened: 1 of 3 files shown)


.github/workflows/hivemq-operator-integration-test.yml
```diff
@@ -50,20 +50,19 @@ jobs:
...
           java-version: '11'
           distribution: 'temurin'
-      - name: Run HiveMQ Operator (legacy) integration tests
-        uses: gradle/gradle-build-action@e2097ccd7e8ed48671dc068ac4efa86d25745b39 # v3
-        env:
-          TEST_PLAN: ${{ matrix.test-plan }}
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@db19848a5fa7950289d3668fb053140cf3028d43 # v3
         with:
-          cache-disabled: true
```
afalhambra-hivemq, line 58: Not sure about this one. By default the cache is enabled using this GitHub actions plugin and to be honest I'm not completely sure why it was disabled in the first place.
If I'm not wrong this is just to cache some of the Gradle tasks, so this shouldn't have any impact on the integration tests other than speeding up the build, right?

afalhambra-hivemq (approved):

.github/workflows/hivemq-operator-integration-test.yml
Donnerbart: Yeah, this should just speed up a consecutive execution by skipping Gradle tasks that have unchanged inputs. Since runners are always freshly started, I see no way how this should create an unwanted cache usage between jobs.

PR #240, opened 2024-05-15
Replace ClusterRole from Helm chart with dependent Role to allow Pod annotation updates when the operator is not running

Donnerbart:
https://hivemq.kanbanize.com/ctrl_board/22/cards/22118/details/

### DIFF
--- a/charts/hivemq-platform-operator/templates/NOTES.txt
+++ b/charts/hivemq-platform-operator/templates/NOTES.txt
@@ -5,7 +5,18 @@ Documentation can be found here: https://docs.hivemq.com/hivemq-platform-operato
 {{- if ne .Values.serviceAccount.create true }}
 
 *** Warning ***
-The service account {{ .Values.serviceAccount.name }} should have correct permissions for the Operator to work correctly.
+The ServiceAccount {{ .Values.serviceAccount.name }} should have sufficient permissions for the Operator to work correctly.
+{{- end }}
+{{- if ne .Values.hivemqPlatformServiceAccount.create true }}
 
+*** Warning ***
+The Operator is not configured to create a ServiceAccount for the HiveMQ Platform Pods.
+Please make sure that you configure a valid custom ServiceAccount in the HiveMQ Platform Helm charts values.yaml (see nodes.serviceAccountName).
+{{- end }}
+{{- if ne .Values.hivemqPlatformServiceAccount.permissions.create true }}
+
+*** Warning ***
+The Operator is not configured to create the RBAC permissions (Role and RoleBinding) for the ServiceAccount of the HiveMQ Platform Pods.
+Please make sure the HiveMQ Platform Helm charts values.yaml configures a ServiceAccount (see nodes.serviceAccountName) with sufficient RBAC permissions to list, get and watch Pods.
 {{- end }}
 -----
--- a/charts/hivemq-platform-operator/templates/rbac.yml
+++ b/charts/hivemq-platform-operator/templates/rbac.yml
@@ -104,8 +104,8 @@ rules:
   - apiGroups:
       - rbac.authorization.k8s.io
     resources:
+      - roles
       - rolebindings
-      - clusterroles
     verbs:
       - get
       - list
@@ -114,19 +114,14 @@ rules:
       - patch
       - update
       - delete
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: "hivemq-platform-operator-pod-resource-role-{{.Release.Name}}"
-rules:
   - apiGroups:
-      - ""
+      - rbac.authorization.k8s.io
     resources:
-      - pods
+      - clusterroles
+      - clusterrolebindings
     verbs:
       - get
-      - patch
-      - update
+      - list
+      - watch
 ---
 {{- end }}
--- a/charts/hivemq-platform-operator/templates/rbac.yml
+++ b/charts/hivemq-platform-operator/templates/rbac.yml
@@ -97,10 +97,12 @@ rules:
       - get
       - list
       - watch
+      {{- if .Values.hivemqPlatformServiceAccount.create }}
       - create
       - patch
       - update
       - delete
+      {{- end }}
   - apiGroups:
       - rbac.authorization.k8s.io
     resources:
@@ -110,10 +112,13 @@ rules:
       - get
       - list
       - watch
+      {{- if .Values.hivemqPlatformServiceAccount.permissions.create }}
       - create
       - patch
       - update
       - delete
+      {{- end }}
+  {{- if .Values.hivemqPlatformServiceAccount.permissions.validate }}
   - apiGroups:
       - rbac.authorization.k8s.io
     resources:
@@ -123,5 +128,6 @@ rules:
       - get
       - list
       - watch
+  {{- end }}
 ---
 {{- end }}
--- a/charts/hivemq-platform-operator/values.yaml
+++ b/charts/hivemq-platform-operator/values.yaml
@@ -10,8 +10,8 @@ image:
   pullPolicy: IfNotPresent
   pullSecretName: ""
 
-# Resources, limits and requests are set to equal values
-# Note: Increase resources depending on how many HiveMQ Platforms this operator manages.
+# Resources, limits and requests are set to equal values.
+# Note: Increase resources depending on how many HiveMQ Platforms the Operator manages.
 resources:
   cpu: 512m
   memory: 512M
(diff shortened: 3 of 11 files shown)


charts/hivemq-platform-operator/templates/rbac.yml
```diff
@@ -104,8 +104,8 @@ rules:
   - apiGroups:
       - rbac.authorization.k8s.io
     resources:
+      - roles
       - rolebindings
-      - clusterroles
```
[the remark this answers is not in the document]
Donnerbart: I think we only need these custom permissions if we want to create/update/delete resources. We should be able to get/list all resources with the `view` ClusterRole that we are binding here: https://github.com/hivemq/helm-charts/blob/master/charts/hivemq-platform-operator/templates/bindings.yml#L10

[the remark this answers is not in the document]
Donnerbart: You might be right, I just looked up that default role and it doesn't include roles and bindings:

> view: Allows read-only access to see most objects in a namespace. It does not allow viewing roles or role bindings.
This role does not allow viewing Secrets, since reading the contents of Secrets enables access to ServiceAccount credentials in the namespace, which would allow API access as any ServiceAccount in the namespace (a form of privilege escalation).

https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles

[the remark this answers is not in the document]
Donnerbart: I guess we need some integration tests in here to check these permissions.
We might have a custom ServiceAccount with Role and RoleBinding or ClusterRole and ClusterRoleBinding. I'm not even sure why the existing `HelmCustomServiceAccountIT` doesn't go into an error state (that custom service account shouldn't have any permissions and there is a check for that).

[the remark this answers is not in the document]
Donnerbart: I've updated the section to configure full CRUD permissions for roles and rolebindings, and get, list, watch for clusterroles and clusterrolebindings. That should be the minimum we need.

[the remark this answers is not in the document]
afalhambra-hivemq: BTW, didn't check out myself, but I'm curious then why the `HelmCustomServiceAccountIT` didn't fail when you removed the `clusterroles` permission?

afalhambra-hivemq (approved):

charts/hivemq-platform-operator/templates/rbac.yml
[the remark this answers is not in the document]
Donnerbart: In general `HelmCustomServiceAccountIT` works because the Platform Operator creates the permissions (`Role` and `RoleBinding`) for the configured `ServiceAccount`, no matter if it's a custom or the default `ServiceAccount`.

I've also added `clusterroles` and `clusterrolebindings` as read permissions.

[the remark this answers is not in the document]
Donnerbart: I just gave it a spin with `platformServiceAccount.permissions.create=false` and no RBAC for:
```yaml
  - apiGroups:
      - rbac.authorization.k8s.io
    resources:
      - clusterroles
      - clusterrolebindings
    verbs:
      - get
      - list
      - watch
```
This results in an `ERROR/INVALID_CUSTOM_RESOURCE_SPEC` state with the message:
```
The custom resource spec is invalid: Failure executing: GET at: https://10.96.0.1:443/apis/rbac.authorization.k8s.io/v1/clusterrolebindings. Message: clusterrolebindings.rbac.authorization.k8s.io is forbidden: User "system:serviceaccount:default:hivemq-platform-operator-test-operator" cannot list resource "clusterrolebindings" in API group "rbac.authorization.k8s.io" at the cluster scope. Received status: Status(apiVersion=v1, code=403, details=StatusDetails(causes=[], group=rbac.authorization.k8s.io, kind=clusterrolebindings, name=null, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=clusterrolebindings.rbac.authorization.k8s.io is forbidden: User "system:serviceaccount:default:hivemq-platform-operator-test-operator" cannot list resource "clusterrolebindings" in API group "rbac.authorization.k8s.io" at the cluster scope, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Forbidden, status=Failure, additionalProperties={}).
```
I think that's fine, since it will only occur if you mess with the Helm chart in an unsupported way.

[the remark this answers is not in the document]
Donnerbart: I gave it another spin with the actual branches and `platformServiceAccount.create=false` and `platformServiceAccount.permissions.create=false`, but activated validation. Using the example K8s resource scripts, I could go step by step through the different error stages of "service account is missing" -> adding service account -> "no permissions" -> adding rolebinding -> "rolebinding roleRef role is missing" -> adding role -> RUNNING.

afalhambra-hivemq (approved):
LGTM!

afalhambra-hivemq:
FYI: Probably there is nothing we want to change there, but in the [NOTES.txt](https://github.com/hivemq/helm-charts/blob/master/charts/hivemq-platform-operator/templates/NOTES.txt#L5-L10) we display a warning when the `serviceAccount` is created externally. Not sure, if you want to add something else to it, or is it fine as is right now.
Just a small heads-up

charts/hivemq-platform-operator/values.yaml
```diff
@@ -42,6 +42,18 @@ serviceAccount:
   # The service account name to be used, or "hivemq-<release name>" if not set.
   name: ""
 
+# Configures the service account and permissions the operator creates for the HiveMQ Platform Pods.
+platformServiceAccount:
```
mhofsche, line 46: I find this a bit confusing. So there are two service accounts, one for the operator to do its work and one for each hivemq platform to access the operator? We should improve the description for both.

I would suggest to find a different name than "platformServiceAccount". How about hivemqPodServiceAccount or hivemqPlatformServiceAccount?

charts/hivemq-platform-operator/values.yaml
```diff
@@ -42,6 +42,18 @@ serviceAccount:
...
+  # Specifies whether a service account should be created for each HiveMQ Platform.
+  create: true
+  # Specifies whether the service account for the HiveMQ Platform should be validated.
```
mhofsche, line 49: Would it make sense to allow a custom service account to be set here just like the above "serviceAccount" setting. Than both types of serviceAccounts are providing the same options.

charts/hivemq-platform-operator/values.yaml
```diff
@@ -42,6 +42,18 @@ serviceAccount:
...
+  validate: true
```
mhofsche, line 50: Is it necessary to make "validate" configurable? What if we always validate no matter what?

charts/hivemq-platform-operator/templates/NOTES.txt
```diff
@@ -5,7 +5,18 @@ Documentation can be found here: https://docs.hivemq.com/operator/latest/kuberne
 {{- if ne .Values.serviceAccount.create true -}}
 
 *** Warning **
-The service account {{ .Values.serviceAccount.name }} should have correct permissions for the Operator to work correctly.
+The ServiceAccount {{ .Values.serviceAccount.name }} should have correct permissions for the Operator to work correctly.
+{{- end }}
+{{- if ne .Values.platformServiceAccount.create true -}}
 
+*** Warning **
+The Operator will not create a ServiceAccount for the HiveMQ Platform Pods.
```
mhofsche, line 13:
```suggestion
The Operator is not configured to create a ServiceAccount for the HiveMQ Platform Pods.
```

charts/hivemq-platform-operator/templates/NOTES.txt
```diff
@@ -5,7 +5,18 @@ Documentation can be found here: https://docs.hivemq.com/operator/latest/kuberne
...
+Please make sure that you configure a valid custom ServiceAccount in the HiveMQ Platform values.yaml (see nodes.serviceAccountName).
+{{- end }}
+{{- if ne .Values.platformServiceAccount.permissions.create true -}}
+
+*** Warning **
+The Operator will not create the Role and RoleBinding for the ServiceAccount of the HiveMQ Platform Pods.
```
mhofsche, line 19:
```suggestion
The Operator is not configured to create the Role and RoleBinding for the ServiceAccount of the HiveMQ Platform Pods.
```

charts/hivemq-platform-operator/templates/NOTES.txt
```diff
@@ -5,7 +5,18 @@ Documentation can be found here: https://docs.hivemq.com/operator/latest/kuberne
...
+Please make sure that the ServiceAccount that is configured in the HiveMQ Platform values.yaml (see nodes.serviceAccountName) has the correct RBAC permissions to list, get and watch Pods.
```
mhofsche, line 20:
```suggestion
Please make sure the HiveMQ Platform values.yaml configures a ServiceAccount (see nodes.serviceAccountName) and has the correct RBAC permissions to list, get and watch Pods.
```

charts/hivemq-platform-operator/values.yaml
Donnerbart: The ServiceAccount for the Platform is for the init app in the Platform Pods, so it can add annotations to the Pod itself.

The Platform is not accessing the operator via K8s. The Platform Pods execute HTTP requests to the operator service to download the init app, but that's a network policy, nothing in the operator deployment.

charts/hivemq-platform-operator/values.yaml
Donnerbart: Then we always need the clusterroles and clusterrolebindings get/list/watch permissions. I want to make them configurable, so customers can limit the permissions the operator needs.

For now I cannot limit the read permissions further, since we still have a dependent resource for ServiceAccount, Role and RoleBinding. But this might change in the future. Then the read permissions for those resource could also be made configurable/optional (when validation is not configured).

charts/hivemq-platform-operator/values.yaml
Donnerbart: I like `hivemqPlatformServiceAccount` better, that since we don't use "pod" in the naming of the HiveMQ Platform Helm chart.

charts/hivemq-platform-operator/values.yaml
Donnerbart: We would need to add another configuration option to the operator then, which would change the default name of the service account for all reconciled platforms. If the platform would configure the service account name, this would then override this setting.

Priority (highest first):
* HiveMQ Platform Helm chart: `nodes.serviceAccount`
* HiveMQ Platform Operator Helm chart: `hivemqPlatformServiceAccount.name`
* HiveMQ Platform Operator code base

Donnerbart: Added `hivemqPlatformServiceAccount.name`

charts/hivemq-platform-operator/values.yaml
Donnerbart: Renamed the `values.yaml` and all usages to `hivemqPlatformServiceAccount`. Also polished the descriptions a bit (and aligned how we spell K8s resource names).

mhofsche (approved):
A few more text suggestions. Otherwise approved.

PR #244, opened 2024-05-21
Add Sidecar containers support

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/22061/details/

### DIFF
--- a/charts/hivemq-platform/values.yaml
+++ b/charts/hivemq-platform/values.yaml
@@ -63,6 +63,14 @@ nodes:
   # The name of the custom service account to use by the Platform pods.
   # If not set or empty, a default service account is created.
   serviceAccountName: ""
+  # Custom containers to be added to the HiveMQ platform pod
+  sidecars: []
+  #  - name: my-sidecar
+  #    image: alpine:latest
+  #    command: ['sh', '-c', 'while true; do echo "logging" >> /opt/hivemq/operator/logs.txt; sleep 30; done']
+  #    volumeMounts:
+  #      - name: operator-init
+  #        mountPath: /opt/hivemq/operator/
 
 # Configuration options for the default provided HiveMQ configuration (config.xml)
 #
--- a/charts/hivemq-platform/values.yaml
+++ b/charts/hivemq-platform/values.yaml
@@ -443,6 +443,7 @@ additionalVolumes: []
 #       It defaults to the "hivemq" container.
 # path: The path configures the directory to which the volume is mounted in the container.
 #       If the directory already exists, then the contents are overwritten!
-# subPath: Optional name for the subPath. If a volume is mounted with subPath, then the contents
-#          of the directory are not overwritten. However, note that changes to a configmap, for example,
-#          mounted via subPath will not be propagated to the container. Therefore, pods have to be restarted.
+# subPath: Optional name for the subPath. The generated volume mount path will be a result of concatenating `path` and the `subPath` values together.
+#       If a volume is mounted with subPath, then the contents of the directory are not overwritten.
+#       However, note that changes to a configmap, for example, mounted via subPath will not be propagated to the container.
+#       Therefore, pods have to be restarted.
--- a/charts/hivemq-platform/values.yaml
+++ b/charts/hivemq-platform/values.yaml
@@ -430,14 +430,19 @@ additionalVolumes: []
 #  - type: configMap
 #    name: configmap-name
 #    mountName: mount-name
+#    containerName: container-name
 #    path: /file/mount/path
 #    subPath: ""
 
 # type: Choose a type of volume that you want to mount.
-# name: Optional name for the secret or the configmap to be mounted
-# mountName: The volume mount name to be used for the StatefulSet Spec.
-# path: The path configures the directory to which the volume is mounted in the container
-#       If the directory already exists then the contents are overwritten!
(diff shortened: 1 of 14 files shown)


afalhambra-hivemq:
Jenkins CI is failing due to the discrepancies between the CRD existing in this `helm-chart` branch (epic branch) and the one existing in the `hivemq-platform-operator` master branch (the `enabled` flag was introduced and merged in the epic branch `epic/platform-operator-1.4.0` but not in `master`).
I think the our Jenkins pipeline does not support multiple repos being in a different branch other than the PR branch or the `master` branch (there is no option for epic branches as such).

@Donnerbart - Should we include this change into `master` as well, or should I just work around this by opening a fake PR on the `hivemq-platform-operator` with the same branch name using the `epic/platform-operator-1.4.0` as the base?. Another potential solution, and I guess ideal, would be to modify the pipeline to allow enter custom base branches (manually from Jenkins?) such as this `epic/platform-operator-1.4.0` branch.

charts/hivemq-platform/values.yaml
```diff
@@ -422,14 +430,19 @@ additionalVolumes: []
...
-#          of the directory are not overwritten. However, note that changes to a configmap for example
+# name: The name used for the Secret, the ConfigMap or the PersistentVolumeClaim to be mounted.
+#       It will be used either as the volume name or as the volume mount name when the `mountName` is not present as well.
+# mountName: Optional volume name and volume mount name to be used for the StatefulSet Spec.
+#       When not present, `name` will be used instead as volume name and as volume mount name.
+# containerName: Optional name for the container on which the volume will be mounted on.
+#       It defaults to the "hivemq" container.
+# path: The path configures the directory to which the volume is mounted in the container.
+#       If the directory already exists, then the contents are overwritten!
+# subPath: Optional name for the subPath. If a volume is mounted with subPath, then the contents
```
Donnerbart, line 446: Maybe we should clarify that the `subPath` is automatically appended to the `path`. See my concerns here: https://github.com/hivemq/hivemq4-documentation/pull/1606#discussion_r1610174542

Otherwise experienced K8s users might configure the "correct" path how K8s expects it.

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/single/HelmSidecarContainerIT.java
```diff
@@ -0,0 +1,50 @@
...
+            assertThat(volumes.stream()) //
+                    .anyMatch(volume -> volume.getName().equals("license-volume") &&
+                            volume.getEmptyDir() != null);
+        });
+        waitForPlatformLog(".*License file license.lic is corrupt.");
+    }
+
+    @Override
+    protected boolean deletePlatformNamespace() {
+        // Cannot delete namespace as Sidecar containers will remain in the pod even if the main container is stopped
```
Donnerbart, line 47: Hmm, doesn't the whole Pod get terminated and sends a SIGTERM or SIGKILL to all containers? I guess your sample sidecars are just not reacting to this, which would leave them running for the whole remaining testsuite? Can we try to fix this?

In the Platform Operator Init App tests I'm using this bash script as a fake broker `run.sh`:
```shell
#!/usr/bin/env bash

terminate() {
  echo "Termination signal received. Exiting gracefully..."
  exit 0
}

# trap the termination signal
trap terminate SIGTERM

# loop to keep the process alive
while true; do
  sleep 1
done
```
This should be stoppable, but otherwise runs in an infinite loop.

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/util/K8sUtil.java
```diff
@@ -382,4 +400,16 @@ public static void assertMqttService(
...
         });
     }
+
+    private static <T extends HasMetadata> @NotNull Resource<T> loadResource(
+            final @NotNull KubernetesClient client,
+            final @NotNull String namespace,
+            final @NotNull String resourceName,
+            final @NotNull Class<T> clazz) {
+        try (final InputStream is = K8sUtil.class.getClassLoader().getResourceAsStream(resourceName)) {
+            return client.resources(clazz).inNamespace(namespace).load(is);
```
Donnerbart, line 410: I'm not 100% sure but this could break in a future Fabric8 version. At least in the Platform Operator the namespace doesn't seem to be applied anymore on the returned resource. I've replaced this with the following code:
```suggestion
            final var resource = client.resources(clazz).load(is);
            assertThat(resource).isNotNull();
            resource.item().getMetadata().setNamespace(namespace);
            return resource;
```

charts/hivemq-platform/values.yaml
afalhambra-hivemq: Updated. Didn't include a small example cause it is becoming quite long description, but I guess this should be enough? let me know your throughts.

Donnerbart (approved):
The new helper functions are scary and awesome at the same time!

Love the improvements, although we had to battle the consul-template setup quite a bit :D

afalhambra-hivemq:
> The new helper functions are scary and awesome at the same time!
> 
> Love the improvements, although we had to battle the consul-template setup quite a bit :D

Yeah, I would say the former more than the latter 😄 . Thanks for the review and troubleshooting

PR #245, opened 2024-05-22
Improve GitHub actions

Donnerbart:
Could be further improved using a composite action:
https://docs.github.com/en/actions/creating-actions/creating-a-composite-action
https://github.blog/changelog/2021-08-25-github-actions-reduce-duplication-with-action-composition/

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -1,6 +1,6 @@
 name: Smoke Test
 on:
-  push:
+  pull_request:
     branches:
       - "**"
   workflow_dispatch:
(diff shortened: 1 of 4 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -23,23 +23,28 @@ jobs:
...
           path: helm-charts
+
       - name: Checkout HiveMQ Platform Operator on the default branch
         uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4
         with:
           repository: hivemq/hivemq-platform-operator
           token: ${{ secrets.JENKINS_TOKEN }}
           path: hivemq-platform-operator
+
       - name: Checkout HiveMQ Platform Operator on the same branch
```
afalhambra-hivemq, line 48: I think we are missing the step here to checkout the HiveMQ Platform Operator on the target branch as well.

Donnerbart: Good catch! It's also needed for the legacy operator job in the same file.

afalhambra-hivemq (approved):
Nice GHA improvements for epics!

PR #251, opened 2024-05-28
Removed `sidecars` and added new top root elements `addtionalContainers` and `additionalInitContainers`

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/22061/details/


charts/hivemq-platform/templates/NOTES.txt
```diff
@@ -1,5 +1,12 @@
 -----
 HiveMQ Platform release "{{.Release.Name}}" installed in namespace "{{.Release.Namespace}}"
 
-Documentation can be found here: https://docs.hivemq.com/operator/latest/kubernetes-operator/platform-operator-for-k8s
+Documentation can be found here: https://docs.hivemq.com/hivemq-platform-operator/index.html
+{{- if .Values.config.overrideInitContainers }}
+
+*** Warning ***
+`config.overrideInitContainers` value is deprecated and it might be removed in future releases.
```
Donnerbart, line 8:
```suggestion
`config.overrideInitContainers` value is deprecated and will be removed in future releases.
```

charts/hivemq-platform/values.schema.json
```diff
@@ -33,6 +33,20 @@
         }
       }
     },
+    "additionalInitContainers": {
```
Donnerbart, line 36: Can we already remove `overrideInitContainers` from the schema? Is this schema used in editors like a `config.xsd`, so customers would directly see that `overrideInitContainers` isn't valid anymore? This could also prevent to guide new customers to only use `additionalInitContainers`.

tests-hivemq-platform-operator/build.gradle.kts
```diff
@@ -81,7 +81,7 @@ val integrationTest by tasks.registering(Test::class) {
         val testPlan = environment["TEST_PLAN"].toString()
         if (testPlan == "Other") {
             systemProperty(
-                "excludeTags", "Upgrade,Extensions,Services1,Services2,CustomConfig,Services,Platform,PodSecurityContext"
+                "excludeTags", "Upgrade,Extensions,CustomConfig,Containers,Services,Platform,PodSecurityContext"
```
Donnerbart, line 84: In the GitHub action I still see `Services1` and `Services2` being used. Why is this just `Services` in here? I assume all `Services1` and `Services2` tests have also the `Services` tag, so this just saves to list all sub-tags? Nice. I think I'll copy this in the platform operator with all the extension tests.

Donnerbart: Nice, I've already copied this over for the operator.

charts/hivemq-platform/tests/hivemq_platform_release_notes_test.yaml
```diff
@@ -0,0 +1,35 @@
...
+    asserts:
+      - equalRaw:
+          value: |
+            -----
+            HiveMQ Platform release "hivemq-platform-release" installed in namespace "hivemq-platform-namespace"
+            
+            Documentation can be found here: https://docs.hivemq.com/hivemq-platform-operator/index.html
+            
+            *** Warning ***
+            `config.overrideInitContainers` value is deprecated and it might be removed in future releases.
```
Donnerbart, line 33:
```suggestion
            `config.overrideInitContainers` value is deprecated and will be removed in future releases.
```

charts/hivemq-platform/values.yaml
```diff
@@ -307,13 +299,34 @@ config:
...
+#    command: ['sh', '-c', 'tail -F /opt/logs.txt']
+#    volumeMounts:
+#      - name: data
+#        mountPath: /opt
+
+# Custom containers to be added to the HiveMQ platform pod
+additionalContainers: []
+#  - name: my-custom-container
+#    image: alpine:latest
+#    command: ['sh', '-c', 'while true; do echo "logging" >> /opt/hivemq/operator/logs.txt; sleep 30; done']
```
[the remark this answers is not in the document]
afalhambra-hivemq: Just wanted to give an example with some volume mounts as well in it, and the `/opt/hivemq/operator/` is one volume already available out of the box. But will change that to something clearer.

charts/hivemq-platform/values.yaml
```diff
@@ -307,13 +299,34 @@ config:
...
   dataHub:
     dataValidationEnabled: false
     behaviorValidationEnabled: false
 
+# Custom init containers to be added to the HiveMQ platform StatefulSet
+additionalInitContainers: []
+#  - name: my-custom-init-container
+#    image: alpine:latest
+#    restartPolicy: Always
+#    command: ['sh', '-c', 'tail -F /opt/logs.txt']
```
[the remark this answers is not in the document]
afalhambra-hivemq: No, it does not. Changing that.

charts/hivemq-platform/values.schema.json
afalhambra-hivemq: Removed from the schema

Donnerbart: As long as you can still use the old configuration (if it's valid), I think it should be okay to remove it from the schema. This way new users don't get the old option suggested anymore (at least I assume there are tools using the schema file also for suggestions and auto-completion like an XSD).

afalhambra-hivemq:
There is something fishy going on with the GitHub runners as per screenshot. They were cancelled after 4mins, re-triggering these couple of jobs again.

image

Donnerbart (approved):
Much cleaner now and allows both types of sidecar configurations. Great job!

mhofsche (approved):

PR #256, opened 2024-06-04
Added VolumeClaimTemplates support for HiveMQ Platform chart

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/22008/details/

### DIFF
--- a/charts/hivemq-platform/templates/hivemq-custom-resource.yml
+++ b/charts/hivemq-platform/templates/hivemq-custom-resource.yml
@@ -186,14 +186,18 @@ spec:
               {{- include "hivemq-platform.get-additional-volumes" . | indent 12 }}
             {{- end }}
           {{- end }}
-          {{- if and (typeIsLike "map[string]interface {}" .Values.podScheduling.affinity) (.Values.podScheduling.affinity) }}
+          {{- with .Values.podScheduling.affinity }}
           affinity:
-            {{- .Values.podScheduling.affinity | toYaml | nindent 12 }}
+            {{- toYaml . | nindent 12 }}
           {{- end }}
-          {{- if and (typeIsLike "[]interface {}" .Values.podScheduling.tolerations) (.Values.podScheduling.tolerations) }}
+          {{- with .Values.podScheduling.tolerations }}
           tolerations:
-            {{- .Values.podScheduling.tolerations | toYaml | nindent 12 }}
+            {{- toYaml . | nindent 12 }}
           {{- end }}
+      {{- with .Values.volumeClaimTemplates }}
+      volumeClaimTemplates:
+        {{- toYaml . | nindent 12 }}
+      {{- end }}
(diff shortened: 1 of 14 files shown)


charts/hivemq-platform/templates/hivemq-custom-resource.yml
```diff
@@ -186,14 +186,18 @@ spec:
               {{- include "hivemq-platform.get-additional-volumes" . | indent 12 }}
             {{- end }}
           {{- end }}
-          {{- if and (typeIsLike "map[string]interface {}" .Values.podScheduling.affinity) (.Values.podScheduling.affinity) }}
+          {{- with .Values.podScheduling.affinity }}
```
Donnerbart, line 189: Just for reference: https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/single/HelmVolumeClaimTemplatesIT.java
```diff
@@ -0,0 +1,106 @@
...
+
+import java.time.Duration;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+
+@Tag("Volumes")
+class HelmVolumeClaimTemplatesIT extends AbstractHelmChartIT {
```
Donnerbart, line 21: We discussed that it would be beneficial to add some asserts that the PersistentVolumes and the VolumeClaimTemplates are working correctly (in comparison to normal Volumes).

So maybe add assertions that each Pod has its own PV and that they are re-mounted when the Pod restarts.

charts/hivemq-platform/templates/hivemq-custom-resource.yml
Donnerbart: This is really nice, much easier to read (once you know how `with` works)!

mhofsche (commented):
Tested with different storage classes and setups. 
We need to document that changes to the volumeClaimTemplate will not take effect if a previous volumeClaimTemplate is present. Changes seem to require manually deleting PVC.

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/single/HelmVolumeClaimTemplatesIT.java
afalhambra-hivemq: done

Donnerbart (approved):
Very nice feature set, also great effort in testing the PVCs properly in the integration test!

mhofsche (approved):

PR #263, opened 2024-06-18
Switch project to Gradle version catalog

afalhambra-hivemq:


tests-hivemq-operator/build.gradle.kts
```diff
@@ -19,24 +19,31 @@ repositories {
...
-    testImplementation("ch.qos.logback:logback-core:${property("logback.version")}")
-    testImplementation("org.slf4j:slf4j-api:${property("slf4j.version")}")
-    testImplementation("org.testcontainers:hivemq:${property("testcontainers.version")}")
+    // Testing
+    testImplementation(libs.assertj)
+    testImplementation(libs.awaitility)
+
+    // JUnit
+    testImplementation(libs.junit.jupiter)
+    testImplementation(libs.junit.jupiter.api)
```
Donnerbart, line 28: I'm pretty sure we don't need the API dependency if we have `junit-jupiter` as dependency. That should be a transitive dependency.

afalhambra-hivemq: good one, thanks

Donnerbart (approved):
Awesome cleanup of the version management!

afalhambra-hivemq:
Re-triggered manually the CI job, it seems the report status was not sent back or the job was not triggered at all.

PR #289, opened 2024-07-18
Supports custom Platform service names and makes metrics service configurable

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/24262/details/


tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/compose/HelmBridgeExtensionIT.java
```diff
@@ -50,13 +48,11 @@ void withBridgeConfiguration_messageBridged() throws Exception {
...
         // deploy chart and wait to be ready
-        installPlatformChartAndWaitToBeRunning("/files/bridge-test-values.yaml");
+        installPlatformChartAndWaitToBeRunning("/files/bridge-values.yaml");
 
         // forward the port from the service
         MqttUtil.execute(client,
-                platformNamespace,
-                MQTT_SERVICE_NAME,
-                MQTT_SERVICE_PORT,
+                platformNamespace, DEFAULT_MQTT_SERVICE_NAME, DEFAULT_MQTT_SERVICE_PORT,
```
Donnerbart, line 55: Nitpick: This file needs to be reformatted.

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/single/CustomClusterDomainNameIT.java
```diff
@@ -25,13 +25,9 @@ class CustomClusterDomainNameIT extends AbstractHelmChartIT {
...
     void whenK8sHasCustomClusterDomainName_thenServicesAreRunning() throws Exception {
-        installPlatformChartAndWaitToBeRunning("/files/mqtt-test-values.yaml");
+        installPlatformChartAndWaitToBeRunning("/files/mqtt-values.yaml");
         K8sUtil.assertMqttService(client, platformNamespace, MQTT_SERVICE_NAME);
-        MqttUtil.execute(client,
-                platformNamespace,
-                MQTT_SERVICE_NAME,
-                MQTT_SERVICE_PORT,
-                MqttUtil.withDefaultPublishSubscribeRunnable());
+        MqttUtil.assertMessages(client, platformNamespace, MQTT_SERVICE_NAME, MQTT_SERVICE_PORT);
```
Donnerbart, line 30: Nitpick: This file needs to be reformatted.

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/util/PrometheusUtil.java
```diff
@@ -0,0 +1,43 @@
...
+            final var request = HttpRequest.newBuilder()
+                    .uri(URI.create("http://localhost:" + portForward.getLocalPort() + path))
+                    .timeout(Duration.ofSeconds(10))
+                    .GET()
+                    .build();
+
+            final var response = client.send(request, HttpResponse.BodyHandlers.ofString());
+            return new TreeMap<>(response.body()
+                    .lines()
+                    .filter(s -> !s.startsWith("#"))
```
Donnerbart, line 37:
```suggestion
                    .filter(s -> !s.isBlank())
                    .filter(s -> !s.startsWith("#"))
```

charts/hivemq-platform/templates/hivemq-custom-resource.yml
```diff
@@ -2,23 +2,24 @@
 {{- $hasLicense := ( include "hivemq-platform.has-license" . ) -}}
 {{- $hasAdditionalVolumes := ( include "hivemq-platform.has-additional-volumes" . ) -}}
 {{- $hasAdditionalVolumeMounts := ( include "hivemq-platform.has-additional-volume-mounts" . ) -}}
-{{- $defaultMetricsPort := ( include "hivemq-platform.metrics-port" . ) -}}
+{{- $metricsPort := ( include "hivemq-platform.metrics-port" . ) -}}
+{{- $hasMetricService := ( include "hivemq-platform.has-default-metrics-service" . ) }}
```
Donnerbart, line 6: I never get these subtle differences, but before the last include line ended with `-}}`, not `}}`. Should this be aligned?
```suggestion
{{- $hasMetricService := ( include "hivemq-platform.has-default-metrics-service" . ) -}}
```

charts/hivemq-platform/templates/hivemq-custom-resource.yml
```diff
@@ -141,15 +142,23 @@ spec:
               image: "{{ .Values.image.repository }}/{{ .Values.image.name }}:{{ .Values.image.tag }}"
               imagePullPolicy: {{ .Values.image.pullPolicy }}
               ports:
+                {{- $containerPortNameList := list }}
+                {{- $containerPortName := "" }}
                 {{- range $key, $val := .Values.services }}
-                {{- if $val.exposed}}
-                - name: {{ include "hivemq-platform.range-service-port-name" . | trimAll "-" | trunc 15 | trimSuffix "-" |  trim }}
-                  containerPort: {{$val.containerPort}}
+                  {{- $containerPortName = (include "hivemq-platform.range-service-port-name" .) | trimAll "-" | trunc 15 | trimSuffix "-" | trim }}
```
Donnerbart, line 148: Can or should we move the `trimAll "-" | trunc 15 | trimSuffix "-" | trim` to the helper method, like we it for `hivemq-platform.range-service-name`? That would make the usage less error prone.

I'm also not sure if we have a check somewhere if the truncation leads to duplicate names. Not sure if this can happen for port names at all, since they are not customizable and we control all the prefixes, right?

charts/hivemq-platform/templates/hivemq-custom-resource.yml
```diff
@@ -233,19 +242,20 @@ spec:
...
     - metadata:
-        name: {{ (include "hivemq-platform.range-service-name" (dict "releaseName" $.Release.Name "type" "metrics" "port" $defaultMetricsPort "containerPort" $defaultMetricsPort)) | trimAll "-" | trunc 63 | trimSuffix "-" |  trim }}
+        name: {{ (include "hivemq-platform.range-service-name" (dict "name" "" "releaseName" $.Release.Name "type" "metrics" "port" $metricsPort "containerPort" $metricsPort)) }}
       spec:
         ports:
-          - name: {{ printf "metrics-%s" $defaultMetricsPort }}
-            targetPort: {{ printf "metrics-%s" $defaultMetricsPort }}
-            port: {{ printf "%d" (int64 $defaultMetricsPort) }}
+          - name: {{ printf "metrics-%s" $metricsPort }}
+            targetPort: {{ printf "metrics-%s" $metricsPort }}
```
Donnerbart, line 256: Do we officially need to add `"metrics"` as supported type to `hivemq-platform.range-service-port-name`? I assume right now this is working since we fallback to `{{- printf "%s-%s" .type (toString .containerPort) }}` in the helper method?

Maybe adding `"metrics`" to the helper method and using it here would make it easier to guarantee that the `StatefulSet` and services configuration are always in sync.

charts/hivemq-platform/tests/hivemq_services_test.yaml
```diff
@@ -184,42 +184,1196 @@ tests:
...
+              ports:
+                - name: cc-8081
+                  targetPort: cc-8081
+                  port: 8081
+              sessionAffinity: ClientIP
+
+  - it: with duplicated containerPorts in enabled services, then services are created successfully
+    set:
+      services:
+        - type: mqtt
```
Donnerbart, line 225: I'm not 100% sure if this scenario makes sense, since the same port should never be able to serve two different service types.

I would see this scenario for the same type though, but then you need to use a custom service name. For example, `mqtt-internal` and `mqtt-external` using the same port (e.g. internal to monitor the same listener, but without load-balancer).

Donnerbart (approved):
This is a great feature. Much respect to improve the helper functions!

charts/hivemq-platform/tests/hivemq_services_test.yaml
afalhambra-hivemq: I think you're right. The main problem I see is because of the listeners defined for the broker config. So with a values like:

```yaml
services:
  - type: mqtt
    exposed: true
    containerPort: 1883
  - type: control-center
    exposed: true
    containerPort: 1883
```

The broker configuration generated will look like:
```xml
...
  <listeners>
    <tcp-listener>
      <port>1883</port>
      <bind-address>0.0.0.0</bind-address>
    </tcp-listener>
  </listeners>
...
  <control-center>
    <listeners>
      <http>
        <port>1883</port>
        <bind-address>0.0.0.0</bind-address>
      </http>
    </listeners>
  </control-center>
  ...
```

Hence, the broker won't start up:

```
ERROR - Could not start TCP Listener on port 1883 and address 0.0.0.0. Is it already in use?
ERROR - Could not bind any listener. Stopping HiveMQ.
```

In a second test with the following values, the same broker configuration will be generated, hence the broker still not coming up:

```yaml
services:
  - type: mqtt
    name: my-mqtt-1
    exposed: true
    containerPort: 1883
  - type: control-center
    name: my-mqtt-2
    exposed: true
    containerPort: 1883
```

In a 3rd test though, I tried with the following values:
```yaml
services:
  - type: mqtt
    name: my-mqtt-1
    exposed: true
    containerPort: 1883
  - type: mqtt
    name: my-mqtt-2
    exposed: true
    containerPort: 1883
```

And the generated broker configuration file has duplicated listeners:
```xml
  <listeners>
    <tcp-listener>
      <port>1883</port>
      <bind-address>0.0.0.0</bind-address>
    </tcp-listener>
    <tcp-listener>
      <port>1883</port>
      <bind-address>0.0.0.0</bind-address>
    </tcp-listener>
  </listeners>
```

So we should probably:
1. Validate there is no combination in a service with duplicated container port and different service type.
2. For same service type, same container port but different service name, it should be valid and the broker configuration should not create a duplicated listener.

charts/hivemq-platform/templates/hivemq-custom-resource.yml
afalhambra-hivemq: Good catch! I'm not 100% sure either whether this is needed in this case, but added that so it's aligned with the rest

charts/hivemq-platform/templates/hivemq-custom-resource.yml
afalhambra-hivemq: Changed the helper method as agreed already

charts/hivemq-platform/templates/hivemq-custom-resource.yml
afalhambra-hivemq: As explained, problem here is that this function is used in two different places with different truncate values. So I could pass this truncate size as a parameter to the helper function but doing that I will need to define as well all the different parameters already used by this function cause doing this, the context is lost (for instance, the .Values are not accessible anymore).

But I think it would be better to have two different functions, even with the duplicated logic (it's not that big), one for the service port name (63 chars max) and another one for the container port name (15 chars max). Will do that.

PR #306, opened 2024-08-14
Adding override option for the resource limits and ephemeral-storage

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/24796/details/


charts/hivemq-platform-operator/templates/NOTES.txt
```diff
@@ -19,4 +19,10 @@ Please make sure that you configure a valid custom ServiceAccount in the HiveMQ
 The Operator is not configured to create the RBAC permissions (Role and RoleBinding) for the ServiceAccount of the HiveMQ Platform Pods.
 Please make sure the HiveMQ Platform Helm charts values.yaml configures a ServiceAccount (see nodes.serviceAccountName) with sufficient RBAC permissions to list, get and watch Pods.
 {{- end }}
+{{- if .Values.resources.overrideLimits }}
+
+*** Warning ***
+`resources.overrideLimits` is meant to be used for testing purposes only.
```
mhofsche, line 25:
```suggestion
`resources.overrideLimits` is not recommended for production use cases and may lead to unpredictable memory and cpu allocation for pods.
```

charts/hivemq-platform-operator/templates/NOTES.txt
```diff
@@ -19,4 +19,10 @@ Please make sure that you configure a valid custom ServiceAccount in the HiveMQ
...
+Overriding resources limits in production environments might lead to unpredictable results.
```
mhofsche, line 26:
```suggestion

```

charts/hivemq-platform/templates/NOTES.txt
```diff
@@ -7,6 +7,11 @@ Documentation can be found here: https://docs.hivemq.com/hivemq-platform-operato
 *** Warning ***
 `config.overrideInitContainers` value is deprecated and will be removed in future releases.
 Please, consider using `additionalInitContainers` value instead.
+{{- end }}
+{{- if .Values.nodes.resources.overrideLimits }}
 
+*** Warning ***
+`nodes.resources.overrideLimits` is meant to be used for testing purposes only.
```
mhofsche, line 14:
```suggestion
`nodes.resources.overrideLimits` is not recommended for production use cases and may lead to unpredictable memory and cpu allocation for pods.
```

charts/hivemq-platform/templates/NOTES.txt
```diff
@@ -7,6 +7,11 @@ Documentation can be found here: https://docs.hivemq.com/hivemq-platform-operato
...
+Overriding resources limits in production environments might lead to unpredictable results.
```
mhofsche, line 15:
```suggestion

```

charts/hivemq-platform-operator/templates/deployment.yml
```diff
@@ -157,11 +157,19 @@ spec:
             timeoutSeconds: 10
           resources:
             limits:
-              cpu: {{.Values.resources.cpu | quote}}
-              memory: {{.Values.resources.memory | quote}}
+              cpu: {{ include "hivemq-platform-operator.resources-limits-cpu" . }}
```
mhofsche, line 160: Can we make CPU limits optional with overrideLimits? There are some suggestions that this may help some use cases: https://kodekloud.com/blog/optimizing-clusters-for-cost-performance-part-2-impact-of-cpu-limits/#:~:text=Tim%20Hockin%2C%20who%20is%20considered,to%20equal%20to%20memory%20requests. 

So if we leave away overrideLimits.cpu then it is not set at all?

charts/hivemq-platform-operator/templates/deployment.yml
```diff
@@ -157,11 +157,19 @@ spec:
...
+              memory: {{ include "hivemq-platform-operator.resources-limits-memory" . }}
+              {{- if and .Values.resources.overrideLimits .Values.resources.overrideLimits.ephemeralStorage }}
+              ephemeral-storage: {{ .Values.resources.overrideLimits.ephemeralStorage | quote }}
```
mhofsche, line 163: Does it make sense to add a helper function for ephemeral-storage here as well?

charts/hivemq-platform/templates/hivemq-custom-resource.yml
```diff
@@ -161,11 +161,19 @@ spec:
                 {{- end }}
               resources:
                 limits:
-                  cpu: {{ .Values.nodes.resources.cpu | quote }}
-                  memory: {{ .Values.nodes.resources.memory | quote }}
+                  cpu: {{ include "hivemq-platform.resources-limits-cpu" . }}
```
mhofsche, line 164: CPU limits optional with overrideLimits logic?

charts/hivemq-platform/templates/hivemq-custom-resource.yml
```diff
@@ -161,11 +161,19 @@ spec:
...
+                  memory: {{ include "hivemq-platform.resources-limits-memory" . }}
+                  {{- if and .Values.nodes.resources.overrideLimits .Values.nodes.resources.overrideLimits.ephemeralStorage }}
+                  ephemeral-storage: {{ .Values.nodes.resources.overrideLimits.ephemeralStorage | quote }}
```
mhofsche, line 167: helper function for ephemeral-storage?

charts/hivemq-platform-operator/templates/deployment.yml
afalhambra-hivemq: I initially added a helper function as well, but the problem here is with the templating language itself... cause in this particular case since this field is optional, we need to wrap the whole `ephemeral-storage` inside the helper function but this creates a lot unnecessary whitespaces and newlines, making the section quite difficult and cumbersome to handle... so I decided to leave this one without any helper function.
I will try to give it another look to see if I can workaround this somehow.

charts/hivemq-platform-operator/templates/deployment.yml
afalhambra-hivemq: Currently, we always set both CPU and memory limits and requests to the default values we have, ie (operator values):

```yaml
resources:
  cpu: 1024m
  memory: 1Gi
```

So if the `overrideLimits.cpu` is not set at all, we are falling back to the default value used in these above original values (i.e. resources.cpu: 1024m).
In order to not set the CPU limits at all, we would then need to modify these new overrideLimits values and explicitly disable it somehow?
```yaml
overrideLimits:
  cpu: 
    enabled: false
```
or enabled:
```yaml
overrideLimits:
  cpu: 
    enabled: true
    value: 1024m
```

charts/hivemq-platform/templates/hivemq-custom-resource.yml
afalhambra-hivemq: Same comment as per https://github.com/hivemq/helm-charts/pull/306/files#r1721421528

charts/hivemq-platform-operator/templates/deployment.yml
afalhambra-hivemq: Managed to do it in a helper function finally 😀

charts/hivemq-platform-operator/templates/deployment.yml
afalhambra-hivemq: As discussed, `overrideLimits` is now modified and it accepts YAML snippets.

charts/hivemq-platform-operator/templates/deployment.yml
```diff
@@ -156,12 +156,25 @@ spec:
             successThreshold: 1
             timeoutSeconds: 10
           resources:
+            {{- if hasKey .Values.resources "overrideLimits" }}
+            {{- with .Values.resources.overrideLimits }}
             limits:
-              cpu: {{.Values.resources.cpu | quote}}
-              memory: {{.Values.resources.memory | quote}}
+              {{- toYaml . | nindent 14 }}
```
mhofsche, line 162: Do we need to quote the yaml snippet, or is this even possible? 

If I configure: 
 resources:
    cpu: 1024m
    memory: 2048M
    overrideLimits:
      cpu: 2048m
      memory: 4096M

I get: 
             resources:
                limits:
                  cpu: 2048m
                  memory: 4096M
                requests:
                  cpu: "1024m"
                  memory: "2048M"

afalhambra-hivemq: There is no need to quote the YAML snippet, usually it's a good practice to quote those values which are string format (cpu, memory, etc), to avoid potential issues with others such as float, integer or boolean. But in this particular case it will converted to a string by default even if you just only type digits.

Even if you ran a `kubectl apply` for a manifest with the following:

```yaml
limits:
  cpu: 2048m
  memory: 4096M
requests:
  cpu: "1024m"
  memory: "2048M"
```

You would only see everything without any quotes in the Kubernetes API
image

afalhambra-hivemq: As I mentioned, for this particular case, it's not really needed, so I'm going to remove the quote function from the template, to get these values aligned.
I tested as well with a running platform to make sure there is no rolling restart out of it because of this change.

mhofsche (approved):

PR #307, opened 2024-08-14
Upgrade kube-prometheus-stack to version 61.*

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/22656/details/

### DIFF
--- a/charts/hivemq-operator/Chart.lock
+++ b/charts/hivemq-operator/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: kube-prometheus-stack
   repository: https://prometheus-community.github.io/helm-charts
-  version: 30.2.0
-digest: sha256:2a5a003a34dcab1e3af23ea9bf4aa3dc8652f3e424704d12e83577aba426e1b9
-generated: "2023-03-16T16:08:27.12931+01:00"
+  version: 61.9.0
+digest: sha256:718571951d52a1a708f5ca5c4ce2508847a5f7453351cfaaddf479124701ba18
+generated: "2024-08-14T14:39:24.136521+02:00"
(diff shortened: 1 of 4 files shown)


charts/hivemq-operator/Chart.lock
```diff
@@ -1,6 +1,6 @@
 dependencies:
 - name: kube-prometheus-stack
   repository: https://prometheus-community.github.io/helm-charts
-  version: 30.2.0
```
mhofsche, line 4: Are there any possible version incompatabilites or upgrade issues between version 30 and 61? Perhaps a serviceagent config change?

afalhambra-hivemq: Yeah, there is actually quite a few changes which require manual intervention. I tried to update a running legacy operator to use this new Prometheus version and it fails (expected as per the [migration guide](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/README.md) for Prometheus):

```
Error: UPGRADE FAILED: cannot patch "operator-prometheus-node-exporter" with kind DaemonSet: DaemonSet.apps "operator-prometheus-node-exporter" is invalid: spec.selector: 
Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"operator", "app.kubernetes.io/name":"prometheus-node-exporter"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable && cannot patch "operator-kube-prometheus-s-node-exporter" with kind PrometheusRule: 
Internal error occurred: failed calling webhook "prometheusrulemutate.monitoring.coreos.com": failed to call webhook: Post "https://operator-kube-prometheus-s-operator.default.svc:443/admission-prometheusrules/validate?timeout=10s": dial tcp 10.111.198.28:443: connect: connection refused
```

Also the Prometheus CRDs should be manually updated as well (Helm only create new ones but do not update existing ones)

PR #317, opened 2024-09-02
Add support for legacy StatefulSet migration

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/24591/details/

### DIFF
--- a/charts/hivemq-platform/values.schema.json
+++ b/charts/hivemq-platform/values.schema.json
@@ -531,13 +531,16 @@
       "if": {
         "properties": {
           "migration": {
+            "type": "object",
+            "required": ["statefulSet"],
             "properties": {
               "statefulSet": {
                 "const": true
               }
             }
           }
-        }
+        },
+        "required": ["migration"]
       },
       "then": {
         "properties": {
--- a/charts/hivemq-platform/values.schema.json
+++ b/charts/hivemq-platform/values.schema.json
@@ -160,6 +160,9 @@
             "type": "object",
             "additionalProperties": true
           },
+          "legacyPortName": {
+            "type": "string"
+          },
           "serviceType": {
             "type" : "string",
             "enum" : [
@@ -226,12 +229,7 @@
               "REQUIRED"
             ]
           }
-        },
-        "required" : [
-          "type",
-          "containerPort",
-          "exposed"
-        ]
+        }
       }
     },
     "extensions" : {
@@ -518,6 +516,56 @@
           "type" : "boolean"
         }
       }
+    },
+    "migration" : {
+      "type" : "object",
+      "properties" : {
+        "statefulSet" : {
+          "type" : "boolean"
+        }
+      }
     }
-  }
+  },
+  "allOf": [
+    {
+      "if": {
+        "properties": {
+          "migration": {
+            "properties": {
+              "statefulSet": {
+                "const": true
+              }
+            }
+          }
+        }
+      },
+      "then": {
+        "properties": {
+          "services": {
+            "items": {
+              "required": [
+                "type",
+                "exposed",
+                "legacyPortName",
+                "containerPort"
+              ]
+            }
+          }
+        }
+      },
+      "else": {
+        "properties": {
+          "services": {
+            "items": {
+              "required": [
+                "type",
+                "exposed",
+                "containerPort"
+              ]
+            }
+          }
+        }
+      }
+    }
+  ]
 }
--- a/charts/hivemq-platform/values.yaml
+++ b/charts/hivemq-platform/values.yaml
@@ -301,14 +301,6 @@ services:
     # The way HiveMQ platform authenticates client certificates. Possible values are `NONE`, `OPTIONAL` and `REQUIRED`
     tlsClientAuthenticationMode: "NONE"
 
-  # Legacy MQTT service configuration
-  # For legacy HiveMQ Operator services, i.e. when the `.migration.statufulSet` flag is enabled, `legacyPortName` is mandatory.
-  - type: mqtt
-    exposed: false
-    containerPort: 8883
-    # The corresponding legacy HiveMQ Operator charts value used in the `hivemq.ports.name` for the service.
-    legacyPortName: "mqtt"
-
 # Configurations and Overrides
 # These configurations provide a convenient option to override entire sections with your customized files,
 # such as the StatefulSet or the HiveMQ configuration (config.xml)
--- a/charts/hivemq-platform/values.yaml
+++ b/charts/hivemq-platform/values.yaml
@@ -230,13 +230,6 @@ services:
     truststorePasswordSecretName: ""
     # The way HiveMQ platform authenticates client certificates. Possible values are `NONE`, `OPTIONAL` and `REQUIRED`
     tlsClientAuthenticationMode: "NONE"
-  # Legacy MQTT service configuration
-  # For legacy HiveMQ Operator services, i.e. when the `.migration.statufulSet` flag is enabled, `legacyPortName` is mandatory.
-  - type: mqtt
-    exposed: false
-    containerPort: 8883
-    # The corresponding legacy HiveMQ Operator charts value used in the `hivemq.ports.name` for the service.
-    legacyPortName: "mqtt"
   # Control Center service configuration
   - type: control-center
     exposed: true
@@ -307,6 +300,14 @@ services:
     truststorePasswordSecretName: ""
     # The way HiveMQ platform authenticates client certificates. Possible values are `NONE`, `OPTIONAL` and `REQUIRED`
     tlsClientAuthenticationMode: "NONE"
+
+  # Legacy MQTT service configuration
+  # For legacy HiveMQ Operator services, i.e. when the `.migration.statufulSet` flag is enabled, `legacyPortName` is mandatory.
+  - type: mqtt
+    exposed: false
+    containerPort: 8883
+    # The corresponding legacy HiveMQ Operator charts value used in the `hivemq.ports.name` for the service.
+    legacyPortName: "mqtt"
 
 # Configurations and Overrides
 # These configurations provide a convenient option to override entire sections with your customized files,
@@ -524,8 +525,3 @@ volumeClaimTemplates: []
 #       requests:
 #         storage: 100Gi
 #     volumeMode: Filesystem
-
-# Migration configuration options for the legacy HiveMQ Operator
-migration:
-  # Enable to migrate from a legacy HiveMQ Operator based on a StatefulSet to the HiveMQ Platform Operator
-  statefulSet: false
--- a/charts/hivemq-platform/values.yaml
+++ b/charts/hivemq-platform/values.yaml
@@ -230,6 +230,13 @@ services:
     truststorePasswordSecretName: ""
     # The way HiveMQ platform authenticates client certificates. Possible values are `NONE`, `OPTIONAL` and `REQUIRED`
     tlsClientAuthenticationMode: "NONE"
+  # Legacy MQTT service configuration
+  # For legacy HiveMQ Operator services, i.e. when the `.migration.statufulSet` flag is enabled, `legacyPortName` is mandatory.
+  - type: mqtt
+    exposed: false
+    containerPort: 8883
+    # The corresponding legacy HiveMQ Operator charts value used in the `hivemq.ports.name` for the service.
+    legacyPortName: "mqtt"
   # Control Center service configuration
   - type: control-center
     exposed: true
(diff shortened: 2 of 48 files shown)


charts/hivemq-platform/values.schema.json
```diff
@@ -518,6 +516,59 @@
...
+    "migration" : {
+      "type" : "object",
+      "properties" : {
+        "statefulSet" : {
+          "type" : "boolean"
+        }
+      }
+    }
+  },
+  "allOf": [
```
Donnerbart, line 529: I didn't know you could have such logic inside a JSON schema. This is impressive!

charts/hivemq-platform/values.yaml
```diff
@@ -301,6 +301,14 @@ services:
     # The way HiveMQ platform authenticates client certificates. Possible values are `NONE`, `OPTIONAL` and `REQUIRED`
     tlsClientAuthenticationMode: "NONE"
 
+  # Legacy MQTT service configuration
```
Donnerbart, line 304: Should we have this in the `values.yaml` if we don't expose the migration switch here? Or should this just go into the migration documentation?

tests-hivemq-platform-operator/build.gradle.kts
```diff
@@ -4,7 +4,7 @@ plugins {
 
 group = "com.hivemq.helmcharts"
 
-val testPlanOtherExcludeTags = "Containers,CustomConfig,Extensions,Platform,PodSecurityContext,ServiceAccount,Services,Upgrade,Volumes"
+val testPlanOtherExcludeTags = "Containers,ClusterDomainName,CustomConfig,Extensions,Platform,PodSecurityContext,ServiceAccount,Services,Upgrade,Volumes"
```
Donnerbart, line 7: Just double checked with the Gradle task which tests are executed in `Other`:
```
Searching for tests in package 'com.hivemq.helmcharts' (includeTags: none) (excludeTags: Containers,ClusterDomainName,CustomConfig,Extensions,Platform,PodSecurityContext,ServiceAccount,Services,Upgrade,Volumes)
Found 7 integration tests:
  com.hivemq.helmcharts.single.HelmManifestsIT
  com.hivemq.helmcharts.single.HelmApplyCrdIT
  com.hivemq.helmcharts.compose.HelmDataHubIT
  com.hivemq.helmcharts.single.HelmInstallRemoteImagesIT
  com.hivemq.helmcharts.compose.HelmRemoveNamespaceIT
  com.hivemq.helmcharts.compose.HelmLegacyStatefulSetMigrationIT
  com.hivemq.helmcharts.single.HelmLicenseSecretIT
```
Looks good to me!

tests-hivemq-platform-operator/src/integrationTest/resources/values/migration-legacy-stateful-set-values.yaml
```diff
@@ -0,0 +1,55 @@
...
+hivemq:
+  nodeCount: 1
+  cpu: 512m
+  controllerTemplate: cluster-stateful-set.yaml
+  ports:
+    - name: "mqtt"
+      port: 1883
+      expose: true
+      patch:
+        - '[{"op":"add","path":"/spec/selector/hivemq.com~1node-offline","value":"false"},{"op":"add","path":"/metadata/annotations","value":{"service.spec.externalTrafficPolicy":"Local"}}]'
```
Donnerbart, line 14: I think the platform CR is missing the customization for these patches. Could be worth to add those and assert them before and after the migration.

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/util/K8sUtil.java
```diff
@@ -336,11 +310,44 @@ public static void executeInPod(
...
+                .findFirst()
+                .orElseThrow();
+    }
+
     /**
      * Returns a Predicate condition based on the state of a Kubernetes resource.
      */
     @SuppressWarnings("unchecked")
-    public static @NotNull Predicate<GenericKubernetesResource> getHiveMQPlatformStatus(final @NotNull String state) {
+    public static @NotNull Predicate<GenericKubernetesResource> getKubernetesResourceStatus(final @NotNull String expectedState) {
```
Donnerbart, line 350: I find this method name a bit misleading, since it will only work for K8s resources that have a `state` field in their status. Without weird coincidences this will only apply to the legacy and platform custom resources.

Maybe `getCustomResourceStateCondition()` is a bit more precise?

charts/hivemq-platform/values.yaml
Donnerbart: @mhofsche What's your opinion on this?

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/compose/HelmLegacyStatefulSetMigrationIT.java
```diff
@@ -0,0 +1,171 @@
...
+        K8sUtil.scaleDeployment(client, operatorNamespace, legacyOperatorDeploymentName, 1);
+        K8sUtil.waitForLegacyOperatorPodStateRunning(client, operatorNamespace, LEGACY_RELEASE_NAME);
+        assertThat(client.pods()
+                .inNamespace(operatorNamespace)
+                .withLabels(legacyOperatorLabels)
+                .list()
+                .getItems()
+                .size()).isEqualTo(1);
+
+        //TODO: We cannot execute `helm uninstall my-release --cascade orphan` - See: https://github.com/helm/helm/issues/13279
```
Donnerbart, line 128: Nitpick, feel free to ignore: I was a bit confused to read the TODO and then a comment that states the opposite. I would change these two lines. So it first says what we really want to, then gives the explanation why we can't and then the workaround.

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/compose/HelmLegacyStatefulSetMigrationIT.java
```diff
@@ -0,0 +1,171 @@
...
+        // we need to uninstall the legacy release with orphan option so the new platform Helm install takes over existing resources
+        //helmChartContainer.uninstallRelease(LEGACY_RELEASE_NAME, operatorNamespace, "--cascade", "orphan");
+
+        // as a workaround we need to delete the Helm secret installed
```
Donnerbart, line 132: A little bit of context would be great. I looked this up, and it seems like Helm v3 creates versioned secrets for every install, upgrade or roll back of a release. That should be the only management resource that is created. The naming schema should be sth. like `sh.helm.release.v1.<release-name>.v<release-version>`.

I guess this is also why you use a label matcher and not a name matcher, so it will remove all possible versions for the release.

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/compose/HelmLegacyStatefulSetMigrationIT.java
```diff
@@ -0,0 +1,171 @@
...
+        K8sUtil.waitForLegacyOperatorPodStateRunning(client, operatorNamespace, LEGACY_RELEASE_NAME);
+        assertThat(client.pods()
+                .inNamespace(operatorNamespace)
+                .withLabels(legacyOperatorLabels)
+                .list()
+                .getItems()
+                .size()).isEqualTo(1);
+
+        //TODO: We cannot execute `helm uninstall my-release --cascade orphan` - See: https://github.com/helm/helm/issues/13279
+        // we need to uninstall the legacy release with orphan option so the new platform Helm install takes over existing resources
```
Donnerbart, line 129: Also a tiny bit of context would be great (without the need to read the linked issue). Maybe something like:
```suggestion
        // we need to uninstall the legacy release with orphan option so the new platform Helm install takes over existing resources
        //TODO: We cannot execute `helm uninstall my-release --cascade orphan`, because this will delete resources
        // without an owner reference, that are still needed - See: https://github.com/helm/helm/issues/13279
```

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/compose/HelmLegacyStatefulSetMigrationIT.java
```diff
@@ -0,0 +1,171 @@
...
+        final var helmSecretLabels = Map.of("owner", "helm", "name", LEGACY_RELEASE_NAME);
+        client.secrets().inNamespace(operatorNamespace).withLabels(helmSecretLabels).withTimeoutInMillis(5000).delete();
+        assertThat(client.secrets()
+                .inNamespace(operatorNamespace)
+                .withLabels(helmSecretLabels)
+                .list()
+                .getItems()).isEmpty();
+
+        // create the corresponding config.xml file in the new ConfigMap for the platform operator
+        K8sUtil.createConfigMap(client, operatorNamespace, "hivemq-migration-config-map.yml");
```
Donnerbart, line 143: Hmm, I have to do this in the operator IT because there is no Helm chart.
But shouldn't the goal of this IT be that a matching broker `config.xml` is automatically created by the Helm chart?

Wouldn't that also be the way customers would approach the migration? They need to transfer their custom settings from the legacy to the platform `values.yaml`. And it's our your job to offer all config options they might want to transfer.

IMO we should not rely on a `config.xml` being provided.

Donnerbart: Is this even used? I cannot see a `--set-file config.overrideHiveMQConfig=hivemq-migration-config-map.xml` anywhere.

Donnerbart: Ahh, it is used because we explicitly don't create a new config:
```yaml
config:
  create: false
  name: "hivemq-configuration"
```

Is there a config option missing to let the Helm chart create the config?

charts/hivemq-platform/values.yaml
mhofsche: Lets keep this out for now. We make it part of the migration description.

tests-hivemq-operator/src/integrationTest/resources/values/ese-extension-values.yaml
```diff
@@ -1,7 +1,3 @@
-global:
```
mhofsche, line 1: This is just a cleanup or?

mhofsche (approved):
Lets keep migration service config out of values.yaml otherwise good to go.

tests-hivemq-operator/src/integrationTest/resources/values/ese-extension-values.yaml
Donnerbart: The commit says `Remove unnecessary pspEnabled value for legacy operator tests`. So yes, just a cleanup of test resources.

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/compose/HelmLegacyStatefulSetMigrationIT.java
afalhambra-hivemq: That's a very valid point, and I believe it's going to be the most painful issue for customers regarding migration.

My understanding here was that it would be the customers who will responsible for providing the right `config.xml` file (based on their own legacy values used, environment variable used for the placeholders, etc.). So I tried to follow the same approach, and at the same time I didn't want to go too deep into using a custom `config.xml` file for that, so I'm using just only the default values.

But you're right, the IT should aim to mimic how customer should try to match all the options from the legacy chart to the new one regarding the `config.xml` file generation and then handle it from there. Specially important when it comes to the listeners defined in a custom `config.xml` as these need to match with the ones defined in values YAML file (`services` section).

> Is there a config option missing to let the Helm chart create the config?

I don't think so. As I mentioned, I tried to use what I thought it would be the most typical use case, but it makes more sense to generate a configuration file based on the values set in a customer values YAML file rather than relying on a provided custom configuration file.

Will change that and create a custom `config.xml` from legacy and its counterpart in the Platform chart.

tests-hivemq-platform-operator/src/integrationTest/resources/values/migration-legacy-stateful-set-values.yaml
afalhambra-hivemq: Good one!

afalhambra-hivemq: As agreed, I will create a new card for customizing services with `sessionAffinity` and assert the annotation only

Donnerbart (approved):
This is awesome work!

PR #322, opened 2024-09-05
Improve support for HiveMQ licenses

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/22981/details/

With this PR we are now providing:
- Support for having an additional HiveMQ broker running side by side with the existing one.
- Support for adding any HiveMQ Enterprise Extension licenses to be part of the same license Secret generated for the HiveMQ broker. Also for extensions, we can have another license running side by side but with a different filename.
- Support for Data Hub license. Also for Data Hub, we can have another license running side by side but with a different filename.


charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,96 @@ restApi:
...
   # To create a new secret with a license, set create=true. Otherwise, false to reuse an existing one.
   name: ""
-  # Inlines the License information as an encoded 64-bytes string.
+  # Inlines the HiveMQ broker license information as an encoded 64-bytes string.
   data: ""
-  # Overrides the License information via file using --set-file license.lic.
+  # Overrides the HiveMQ broker license information via file using --set-file license.lic.
   overrideLicense: ""
+  # HiveMQ Enterprise extension licenses to include as part of the same Kubernetes Secret license.
+  extensions: {}
```
afalhambra-hivemq, line 103: A couple of important to notes regarding the implementation details:

1. Since the HiveMQ broker does not support checking extension licenses in subfolder (default license folder is always `/license`), we cannot mount multiple volumes so the same Secret used for the license broker needs to be used here (`subPath` is not an option here as change there won't be reflected on runtime for the broker to reload it).
2. The existing `extensions` values cannot be used to include a new sub-value called `license` or something like that. This is because the type of this `extensions` value is `array` and thus it makes impossible to use the existing feature for the license to inline a file license into the value, i.e. `--set-file extensions.hivemq-kafka-extension.license.elic`. For this to work, we would need to switch it from `array` to `object` as it is already defined under the new element `licenses.extensions` but unfortunately this would be a breaking change. So for that I created a new value `license.extensions` as `object` so this can be done.

mhofsche (approved):
Tested with 4 licenses.

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,103 @@ restApi:
...
+# Configures the HiveMQ Platform license information.
 license:
   # Creates a default HiveMQ Platform license as a Kubernetes Secret.
   create: false
-  # The name of a secret with the license information.
+  # The name of a Kubernetes Secret with the HiveMQ Platform license information.
   # To create a new secret with a license, set create=true. Otherwise, false to reuse an existing one.
   name: ""
-  # Inlines the License information as an encoded 64-bytes string.
+  # Inlines the HiveMQ broker license information as an encoded 64-bytes string.
```
mhofsche, line 98:
```suggestion
  # Inlines the HiveMQ Platform license information as an encoded 64-bytes string.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,103 @@ restApi:
...
   data: ""
-  # Overrides the License information via file using --set-file license.lic.
+  # Overrides the HiveMQ broker license information via file using --set-file license.lic.
```
mhofsche, line 100:
```suggestion
  # Overrides the HiveMQ Platform license information via file using --set-file license.lic.
```

charts/hivemq-platform/values.yaml
afalhambra-hivemq: Superseded already by the new rework.

charts/hivemq-platform/values.yaml
afalhambra-hivemq: Superseded already by the new rework.

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
   # Enables or disables authentication and authorization.
   authEnabled: false
```
mhofsche, line 90: In general we should align to the documentation here. 
The "main" license is just called "HiveMQ license", so I guess we should stick to that. I will try a set of suggestions for this. Let me know what you think.

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
 
-# Configures the HiveMQ License information.
+# Configures the HiveMQ Platform license information.
```
mhofsche, line 91:
```suggestion
# Configures the HiveMQ license information.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
 license:
   # Creates a default HiveMQ Platform license as a Kubernetes Secret.
```
mhofsche, line 93:
```suggestion
  # Creates a default HiveMQ license as a Kubernetes Secret.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
   create: false
-  # The name of a secret with the license information.
+  # The name of a Kubernetes Secret with the HiveMQ Platform license information.
```
mhofsche, line 95:
```suggestion
  # The name of a Kubernetes Secret with the HiveMQ license information.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
   # To create a new secret with a license, set create=true. Otherwise, false to reuse an existing one.
```
mhofsche, line 96:
```suggestion
  # To create a new Kubernetes Secret with a license, set create=true. Otherwise, false to reuse an existing Secret.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
   name: ""
-  # Inlines the License information as an encoded 64-bytes string.
+  # Inlines the HiveMQ broker license information as an encoded 64-bytes string.
```
mhofsche, line 98:
```suggestion
  # Inlines the HiveMQ license information as an encoded 64-bytes string.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
   data: ""
-  # Overrides the License information via file using --set-file license.lic.
+  # Overrides the HiveMQ broker license information via file using --set-file license.lic.
```
mhofsche, line 100:
```suggestion
  # Overrides the HiveMQ license information via file using --set-file license.lic.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
   overrideLicense: ""
+  # Additional HiveMQ broker licenses to include as .lic files of the same Kubernetes Secret license.
```
mhofsche, line 102:
```suggestion
  # Add additional HiveMQ licenses to the Kubernetes Secret as .lic files.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+  additionalBroker: {}
```
mhofsche, line 103:
```suggestion
  additionalLicenses: {}
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    ## Defines the name of the HiveMQ broker license in the Kubernetes secret (the .lic suffix will be added automatically).
```
mhofsche, line 104:
```suggestion
    ## Defines the name of the HiveMQ license in the Kubernetes Secret (the .lic suffix will be added automatically).
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #<broker-license-name>:
```
mhofsche, line 105:
```suggestion
    #license-name:
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #  # Inlines the HiveMQ broker license information as an encoded 64-bytes string.
```
mhofsche, line 106:
```suggestion
    #  # Inlines the HiveMQ license information as an encoded 64-bytes string.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #  data: ""
+    #  # Overrides the HiveMQ broker license information via file using --set-file license.lic.
```
mhofsche, line 108:
```suggestion
    #  # Overrides the HiveMQ license information via file using --set-file license.lic.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #  overrideLicense: ""
+  # HiveMQ Enterprise Extension licenses to include as part of the same Kubernetes Secret license.
```
mhofsche, line 110:
```suggestion

  # HiveMQ Enterprise Extension licenses to include as part of the same Kubernetes Secret. Add required licenses to the list.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+  extensions: {}
+    ## Defines the name of the HiveMQ Enterprise Extension license in the Kubernetes secret (the .elic suffix will be added automatically).
```
mhofsche, line 112:
```suggestion
    ## Defines the name of the HiveMQ Enterprise Extension license in the Kubernetes Secret (the .elic suffix will be added automatically).
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #<extension-license-name>:
```
mhofsche, line 113:
```suggestion
    #extension-license-name:
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #  # Inlines the HiveMQ Enterprise Extension license information as an encoded 64-bytes string.
+    #  data: ""
+    #  # Overrides the HiveMQ Enterprise Extension license information via file using --set-file license.elic.
+    #  overrideLicense: ""
+  # HiveMQ Data Hub licenses to include as part of the same Kubernetes Secret license.
```
mhofsche, line 118:
```suggestion
  # HiveMQ Data Hub licenses to include as part of the same Kubernetes Secret. 
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+  dataHub: {}
+    ## Defines the name of the HiveMQ Data Hub license in the Kubernetes secret (the .plic suffix will be added automatically).
```
mhofsche, line 120:
```suggestion
    ## Defines the name of the HiveMQ Data Hub license in the Kubernetes Secret (the .plic suffix will be added automatically).
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #<data-hub-license-name>:
```
mhofsche, line 121:
```suggestion
    #data-hub-license-name:
```

charts/hivemq-platform/values.schema.json
```diff
@@ -302,17 +310,26 @@
     "license" : {
       "type" : "object",
       "properties" : {
+        "additionalBroker" : {
```
mhofsche, line 313:
```suggestion
        "additionalLicense" : {
```

charts/hivemq-platform/values.yaml
afalhambra-hivemq: In this particular case, I changed it like this cause this whole `license` section includes all the different licenses for the entire platform. Only `HiveMQ license` would be misleading IMO - Maybe something like below, what do you think?

```yaml
Configures all HiveMQ licenses information.
```

charts/hivemq-platform/values.yaml
afalhambra-hivemq: Not so sure here regarding the removal of the minor and greater than symbols.... In all of our examples we have commented out in the values.yaml file, these are existing values defined in the JSON schema, so customer can only use them with that name. 
In this particular case though, it's different cause there is no `license-name` value available as such as the customer can choose whatever name they like. So I think the most conventional way of defining this is by using `<>` or something to clearly denote that this can be any value.
For now, I will remove the `broker` prefix. 
WDYT?

charts/hivemq-platform/values.yaml
afalhambra-hivemq: same comment as the `license-name`. Let me know what you think

charts/hivemq-platform/values.yaml
afalhambra-hivemq: same comment as the `license-name`. Let me know what you think

charts/hivemq-platform/values.yaml
afalhambra-hivemq: Yeah, I agree in most of the comments. Have a look.

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
-  # Creates a default HiveMQ Platform license as a Kubernetes Secret.
+  # Creates a default HiveMQ license as a Kubernetes Secret.
   create: false
-  # The name of a secret with the license information.
-  # To create a new secret with a license, set create=true. Otherwise, false to reuse an existing one.
+  # The name of a Kubernetes Secret with the HiveMQ license information.
+  # To create a new Kubernetes Secret with a license, set create=true. Otherwise, false to reuse an existing Secret.
   name: ""
-  # Inlines the License information as an encoded 64-bytes string.
+  # Inlines the HiveMQ license information as an encoded 64-bytes string.
```
mhofsche, line 98:
```suggestion
  # Inlines the main HiveMQ license as an encoded 64-bytes string.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
   data: ""
-  # Overrides the License information via file using --set-file license.lic.
+  # Overrides the HiveMQ license information via file using --set-file license.lic.
```
mhofsche, line 100:
```suggestion
  # Overrides the main HiveMQ license via file using --set-file license.lic.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
   overrideLicense: ""
+  # Adds additional HiveMQ licenses to the Kubernetes Secret as .lic files.
```
mhofsche, line 102:
```suggestion

  # Adds additional main HiveMQ licenses to the Kubernetes Secret as .lic files if needed.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+  additionalLicenses: {}
+    ## Defines the name of the HiveMQ license in the Kubernetes Secret (the .lic suffix will be added automatically).
```
mhofsche, line 104:
```suggestion
    ## Defines the name of the additional main HiveMQ license (the .lic suffix will be added automatically).
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #<license-name>:
```
mhofsche, line 105:
```suggestion
    #license-name:
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #  # Inlines the HiveMQ license information as an encoded 64-bytes string.
```
mhofsche, line 106:
```suggestion
    #  # Inlines the additional main HiveMQ license as an encoded 64-bytes string.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #  data: ""
+    #  # Overrides the HiveMQ license information via file using --set-file license.lic.
```
mhofsche, line 108:
```suggestion
    #  # Overrides the additional main HiveMQ license via file using --set-file additional-license.lic.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #  overrideLicense: ""
+  # HiveMQ Enterprise Extension licenses to include as part of the same Kubernetes Secret. Add required licenses to the list.
```
mhofsche, line 110:
```suggestion
  # HiveMQ Enterprise Extension licenses to include as part of the Kubernetes Secret. Add your required licenses to the list.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+  extensions: {}
+    ## Defines the name of the HiveMQ Enterprise Extension license in the Kubernetes Secret (the .elic suffix will be added automatically).
```
mhofsche, line 112: We should provide a commented out example for the kafka extension

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #<extension-license-name>:
+    #  # Inlines the HiveMQ Enterprise Extension license information as an encoded 64-bytes string.
```
mhofsche, line 114:
```suggestion
    #  # Inlines the HiveMQ Enterprise Extension license as an encoded 64-bytes string.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #  data: ""
+    #  # Overrides the HiveMQ Enterprise Extension license information via file using --set-file license.elic.
```
mhofsche, line 116:
```suggestion
    #  # Overrides the HiveMQ Enterprise Extension license via file using --set-file license.elic.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #  overrideLicense: ""
+  # HiveMQ Data Hub licenses to include as part of the same Kubernetes Secret.
```
mhofsche, line 118:
```suggestion
  # HiveMQ Data Hub licenses to include as part of the Kubernetes Secret.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+  dataHub: {}
+    ## Defines the name of the HiveMQ Data Hub license in the Kubernetes Secret (the .plic suffix will be added automatically).
+    #<data-hub-license-name>:
+    #  # Inlines the HiveMQ Data Hub license information as an encoded 64-bytes string.
```
mhofsche, line 122:
```suggestion
    #  # Inlines the HiveMQ Data Hub license as an encoded 64-bytes string.
```

charts/hivemq-platform/values.yaml
```diff
@@ -88,17 +88,41 @@ restApi:
...
+    #  data: ""
+    #  # Overrides the HiveMQ Data Hub license information via file using --set-file license.plic.
```
mhofsche, line 124:
```suggestion
    #  # Overrides the HiveMQ Data Hub license via file using --set-file license.plic.
```

mhofsche (approved):

charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -475,6 +474,82 @@ Usage: {{- include "hivemq-platform.validate-additional-volumes" . }}
 {{- end -}}
 {{- end -}}
 
+{{/*
+Validates the broker license, additional broker licenses, the enterprise extension licenses and the Data Hub license so:
+ - When the `license.create` value is enabled, a least one license content is defined.
```
Donnerbart, line 479:
```suggestion
 - When the `license.create` value is enabled, at least one license content is defined.
```

charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -475,6 +474,82 @@ Usage: {{- include "hivemq-platform.validate-additional-volumes" . }}
...
+{{/*
+Validates the licenses content within a range function:
+ - At least either `data` or `overrideLicense` is not empty.
+ - Only one of `data` or `overrideLicense` are set, and not both.
+Usage: {{ include "hivemq-platform.validate-license-content" (dict "licenses" .licenses "licenseLabel" "HiveMQ Enterprise Extension") }}
+*/}}
+{{- define "hivemq-platform.validate-license-content" -}}
+{{- $label := .licenseLabel -}}
+{{- range $licenseName, $license := .licenses -}}
+    {{- if and (not $license.data) (not $license.overrideLicense) -}}
```
Donnerbart, line 509: I think the checks in this method are overlapping a bit. For example, when I configure just the key `foobar` it will print the `cannot be empty` error message, although the `Invalid values` would be more fitting.

charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -475,6 +474,82 @@ Usage: {{- include "hivemq-platform.validate-additional-volumes" . }}
...
+        {{- fail (printf "%s '%s' license content cannot be empty. Please, use either `data` or `overrideLicense` values" $label $licenseName) }}
+    {{- end -}}
+    {{- if and (gt (len $license) 1) (hasKey $license "data") (hasKey $license "overrideLicense") -}}
+        {{- fail (printf "Both `data` and `overrideLicense` values are set for the %s '%s' license content. Please, use only one of them" $label $licenseName) }}
+    {{- else if gt (len $license) 1 -}}
+        {{- fail (printf "Invalid values for setting a license content. Only `data` or `overrideLicense` values are allowed") }}
```
Donnerbart, line 515: The error message doesn't use the `$label` and `$licenseName`

charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -475,6 +474,82 @@ Usage: {{- include "hivemq-platform.validate-additional-volumes" . }}
...
+{{- end -}}
+
+{{/*
+Validates there is no "license.lic" defined in the additional broker license values, so it doesn't clash with the default broker name.
+Usage: {{ include "hivemq-platform.validate-duplicated-additional-broker-license" . }}
+*/}}
+{{- define "hivemq-platform.validate-duplicated-additional-broker-license" -}}
+{{- if or .Values.license.data .Values.license.overrideLicense -}}
+    {{- range $licenseName, $license := .Values.license.additionalLicenses -}}
+      {{- $secretEntry := printf "%s%s" $licenseName ".lic" -}}
```
Donnerbart, line 527: Can't we directly check `if eq $licenseName "license"` here?

charts/hivemq-platform/templates/hivemq-license.yml
```diff
@@ -1,19 +1,19 @@
...
     {{ . | nindent 4 }}
     {{ end }}
-  {{- else }}
-    {{- fail ("HiveMQ Platform license content cannot be empty") }}
-  {{- end }}
   {{- end }}
-kind: Secret
-metadata:
-  name: {{ .Values.license.name | default (include "hivemq-platform.default-license-name" .) }}
+  {{- include "hivemq-platform.generate-licenses-content" (dict "licenses" .Values.license.additionalLicenses "licenseExtension" ".lic") }}
```
Donnerbart, line 16: I really like the abstraction on this level! If there will ever be another license type, this will be very easy to add.

charts/hivemq-platform/templates/_helpers.tpl
afalhambra-hivemq: On line 513, it uses both `$label` and `$licenseName`: 
```
...the %s '%s'...
```

Or not sure if you're referring to line 515 where there is no variable used

charts/hivemq-platform/templates/_helpers.tpl
afalhambra-hivemq: Sorry, having a hard time now with GitHub... I guess you're referring to line 528.
And instead of:
```
{{- if eq $secretEntry "license.lic" -}}
```

Just use:

```
if eq $licenseName "license"
```

If so, yes, we can only check that. No need to include the extension of the file as it will always be `.lic`.

charts/hivemq-platform/templates/_helpers.tpl
Donnerbart: The comment was for the `Invalid values for setting a license content` error message.

charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -475,6 +474,80 @@ Usage: {{- include "hivemq-platform.validate-additional-volumes" . }}
...
+Validates the licenses content within a range function:
+ - At least either `data` or `overrideLicense` is not empty.
+ - Only one of `data` or `overrideLicense` are set, and not both.
+Usage: {{ include "hivemq-platform.validate-license-content" (dict "licenses" .licenses "licenseLabel" "HiveMQ Enterprise Extension") }}
+*/}}
+{{- define "hivemq-platform.validate-license-content" -}}
+{{- $label := .licenseLabel -}}
+{{- range $licenseName, $license := .licenses -}}
+    {{- if and (not (hasKey $license "data")) (not (hasKey $license "overrideLicense")) -}}
+        {{- fail (printf "Invalid values for setting a license content. Only `data` or `overrideLicense` values are allowed") }}
```
Donnerbart, line 510: This error message doesn't use `$label $licenseName`, so it's hard associate the wrong configuration from the error message.

charts/hivemq-platform/templates/_helpers.tpl
Donnerbart: Yes, I think the `{{- $secretEntry := printf "%s%s" $licenseName ".lic" -}}` is not needed and the code can be simplified.

Donnerbart: The current source code is implemented like this, will resolve.

Donnerbart (approved):
This was a long journey, but the result is really amazing, very flexible and extendable for new licence types.

PR #336, opened 2024-09-16
Include PROXY protocol configuration for MQTT and WebSocket listeners

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/25811/details/


charts/hivemq-platform/values.yaml
```diff
@@ -192,6 +192,9 @@ services:
     annotations: {}
     # Labels to add to the service
     labels: {}
+    # Enables PROXY protocol for a HiveMQ listener. Only applicable to MQTT and WebSocket services.
+    # See https://docs.hivemq.com/hivemq/latest/user-guide/proxy-protocol.html
+    # proxyProtocol: false
```
mhofsche, line 197:
```suggestion
    # hivemqProxyProtocol: false
```

We should make sure this is related to HiveMQ config.

PR #337, opened 2024-09-20
Fix Service not being created when service account is disabled

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/25897/details/


Donnerbart (approved):
Nice cleanup!

PR #341, opened 2024-09-24
Update version to 4.33.0

SgtSilvio:


manifests/hivemq-operator/operator/operator-deployment.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: hivemq-operator
+  name: hivemq-operator-operator
```
SgtSilvio, line 6: @Donnerbart @afalhambra-hivemq are these changes intended? They were automatically created while running `./gradlew updateAllPlatformChartVersions -PappVersion=4.33.0`

afalhambra-hivemq: No, definitely not. I will have a look to see what's going wrong here. But for now, please revert the changes in this file where the suffix `-operator` is added

ymengesha (approved):

manifests/hivemq-operator/operator/operator-deployment.yaml
ymengesha: @afalhambra-hivemq we have been seeing some silent `sed` warnings for a while now. 
```
> Task :updateOperatorChartVersion
==================================================
Generating HiveMQ Operator (legacy) Manifest files
==================================================
Create manifests directory
Update HiveMQ Operator dependencies
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "hivemq" chart repository
...Successfully got an update from the "prometheus" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 1 charts
Downloading kube-prometheus-stack from repo https://prometheus-community.github.io/helm-charts
Deleting outdated charts
Create HiveMQ Operator Templates
Flatten directory structure
sed: 1: "hivemq-operator/operato ...": extra characters at the end of h command
```
I have looked into this before and it was related to the difference of the options between the bsd and gnu versions. In Mac, the default one is the BSD one and the correct `sed` command would be: `sed -i '' ...` but this would probably break stuff for @Donnerbart as he is on linux and probably has the gnu version. So, in general, the `-i` option is the culprit. Maybe we can use a different command to do the replacements instead of supporting both.

afalhambra-hivemq: yeah, actually there was an issue with a recent change that kept the `sed -i` command leading to not working properly on MacOS but on Linux.
We fixed that already [here](https://github.com/hivemq/helm-charts/pull/342).

It seems the only valid approach is to use `sed -i.bak` and the delete the `.bak` files afterwards.

PR #343, opened 2024-09-24
Support Container Security Context

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/25915/details/

### DIFF
--- a/charts/hivemq-platform/values.schema.json
+++ b/charts/hivemq-platform/values.schema.json
@@ -132,7 +132,7 @@
       }
     },
     "metrics" : {
-      "type" : "object",
+      "type": "object",
       "properties" : {
         "enabled" : {
           "type" : "boolean"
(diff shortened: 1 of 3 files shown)


charts/hivemq-platform/values.schema.json
```diff
@@ -15,7 +15,11 @@
           "type" : "string"
         },
         "overrideStatefulSet" : {
-          "type" : "string"
+          "type": "string",
+          "contentMediaType": "application/json",
+          "contentSchema": {
+            "$ref": "https://raw.githubusercontent.com/kubernetes/kubernetes/v1.31.1/api/openapi-spec/swagger.json#/definitions/io.k8s.api.apps.v1.StatefulSet"
```
mhofsche, line 21: What happens if we have a non-compatible update for one of these definitions between lets say k8s 1.25 and 1.31. If the customer still runs k8s 1.25 that should be fine. But we would reject that or?

afalhambra-hivemq: Already agreed to discuss this later on.

PR #367, opened 2024-10-21
Fix environment variables with dotted names for HiveMQ Platform Operator

Donnerbart:
https://hivemq.kanbanize.com/ctrl_board/22/cards/27101/details/

### DIFF
--- a/charts/hivemq-platform-operator/templates/deployment.yml
+++ b/charts/hivemq-platform-operator/templates/deployment.yml
@@ -51,64 +51,64 @@ spec:
               value: {{.Values.javaOpts | quote }}
             {{- /* `nil` check needed for backward compatibility */ -}}
             {{- if and (ne .Values.selector "nil") (.Values.selector) }}
-            - name: quarkus.operator-sdk.controllers.hivemq-controller.selector
+            - name: QUARKUS_OPERATOR_SDK_CONTROLLERS__HIVEMQ_CONTROLLER__SELECTOR
               value: operator={{ .Values.selector }}
             {{- end }}
             - name: KUBERNETES_NAMESPACE
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
-            - name: hivemq.platform.operator.release.name
+            - name: HIVEMQ_PLATFORM_OPERATOR_RELEASE_NAME
               value: "{{ .Release.Name }}"
-            - name: hivemq.platform.operator.crd.apply
+            - name: HIVEMQ_PLATFORM_OPERATOR_CRD_APPLY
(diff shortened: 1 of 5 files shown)


afalhambra-hivemq:
We have to change the target branch in this PR.

charts/hivemq-platform-operator/templates/deployment.yml
```diff
@@ -51,64 +51,64 @@ spec:
...
+            - name: HIVEMQ_PLATFORM_OPERATOR_RELEASE_NAME
               value: "{{ .Release.Name }}"
-            - name: hivemq.platform.operator.crd.apply
+            - name: HIVEMQ_PLATFORM_OPERATOR_CRD_APPLY
               value: "{{ .Values.crd.apply }}"
-            - name: hivemq.platform.operator.crd.wait-until-ready
+            - name: HIVEMQ_PLATFORM_OPERATOR_CRD_WAIT_UNTIL_READY
               value: "{{ .Values.crd.waitReady }}"
-            - name: hivemq.platform.operator.crd.wait-until-ready.timeout
+            - name: HIVEMQ_PLATFORM_OPERATOR_CRD_WAIT_UNTIL_READY_TIMEOUT
```
afalhambra-hivemq, line 67: Wasn't sure about these ones with a hyphen, but it seems it's resolved as well without any issues.
It seems that in case that both env var `FOO_BAR_BAZ=VALUE` and an application property `foo.bar-baz=default` exist, the result will be: `foo.bar-baz=value` 
https://smallrye.io/smallrye-config/Main/config/environment-variables/

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/single/CustomOperatorSelectorIT.java
```diff
@@ -0,0 +1,46 @@
...
+class CustomOperatorSelectorIT extends AbstractHelmChartIT {
+
+    @Override
+    protected boolean installPlatformOperatorChart() {
+        return false;
+    }
+
+    @Test
+    @Timeout(value = 5, unit = TimeUnit.MINUTES)
+    void whenK8sHasCustomClusterDomainName_thenServicesAreRunning() throws Exception {
```
afalhambra-hivemq, line 23: Copy&Paste leftover.

afalhambra-hivemq (approved):
LGTM!

charts/hivemq-platform-operator/templates/deployment.yml
Donnerbart: I've fixed the naming to be unambiguous regarding hyphens.

PR #399, opened 2024-11-27
Add ServiceMonitor resources for both HiveMQ Platform Operator Helm charts and HiveMQ Platform Helm charts

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/27143/details/

Edit: Haven't added the new matrix tags to the branch protection rules, as I would like to get first the approval. Will include that and re-trigger the `verify` job afterwards.


charts/hivemq-platform-operator/tests/hivemq_operator_deployment_test.yaml
```diff
@@ -349,8 +358,8 @@ tests:
         runAsUser: 185
     asserts:
       - failedTemplate:
-          errorMessage: |-
-            `runAsNonRoot` is set to `false` but `runAsUser` is not set to `0` (root)
+          errorPattern: |-
```
Donnerbart, line 361: Stupid question: Is there a reason to use `|-` here? I've seen that you refactored a couple of the `errorPattern` into a single line and the majority seems to be like that.

charts/hivemq-platform-operator/files/grafana-dashboard.json
```diff
@@ -0,0 +1,2701 @@
...
+    ]
+  },
+  "time": {
+    "from": "now-30m",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "",
+  "title": "HiveMQ Platform Operator (Prometheus)",
+  "uid": "pmawgrlGk__",
```
Donnerbart, line 2698: I think this is still the `uid` from the original dashboard.

Donnerbart: I can't find this in the JSON, but also the top-left application dropdown didn't work in my operator-only tests. Also the JOSDK and our custom metrics are probably still missing. The missing metrics might be out-of-scope for now, but we should at least have a look at the application selector.

charts/hivemq-platform-operator/tests/hivemq_operator_deployment_test.yaml
afalhambra-hivemq: Not a stupid question! 
The only reason to use `|-` is because this unittest plugin does not work when the error message to match starts with some special character like ``` in this case. If you see, the `|-` is only applied to those starting with such character. It will fail with:
```
 FAIL   charts/hivemq-platform-operator/tests/hivemq_operator_deployment_test.yaml
        - Execution Error: 
                yaml: line 362: found character that cannot start any token
```
Actually, I'm going to raise this an issue on the Helm unittest plugin.
Another option would be to change the error message in our case.

afalhambra-hivemq: https://github.com/helm-unittest/helm-unittest/issues/499

charts/hivemq-platform-operator/values.yaml
```diff
@@ -164,3 +164,32 @@ crd:
...
+    # Overrides the default HiveMQ Platform Grafana Operator dashboard content via file using --set-file grafana-dashboard.json.
+    overrideDashboard: ""
+  serviceMonitor:
+    # Namespace where the ServiceMonitor will be deployed to. Defaults to use the same namespace where the HiveMQ Platform Operator Helm chart is deployed to.
+    # namespace: monitoring
+    # Helm chart release name of the Prometheus Monitoring stack available in the Kubernetes cluster.
+    # This allows the Prometheus operator to discover the HiveMQ Platform Operator ServiceMonitor. Defaults to the HiveMQ Platform Operator Helm chart release name.
+    # releaseName: monitoring-stack
+    # Defines how often Prometheus should scrape metrics from the endpoints defined in the ServiceMonitor resource.
+    # Format is based on the regular expression `(0|(([0-9]+)y)?(([0-9]+)w)?(([0-9]+)d)?(([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?(([0-9]+)ms)?)`. For example 1d, 1h30m, 5m, 10s.
```
mhofsche, line 191: That regex is a bit much here.

charts/hivemq-platform-operator/values.yaml
```diff
@@ -164,3 +164,32 @@ crd:
...
+
+# Monitoring configuration options
+monitoring:
+  # When enabled, deploys a ServiceMonitor resource for the Operator along with a Configmap for a Grafana dashboard.
+  enabled: false
+  # Dashboard configuration options.
+  dashboard:
+    # Creates a default HiveMQ Platform Operator Grafana dashboard ConfigMap.
+    create: true
+    # Namespace where the Grafana dashboard ConfigMap will be deployed to. Defaults to use the same namespace where the HiveMQ Platform Operator chart is deployed to.
```
mhofsche, line 176:
```suggestion
    # Namespace where the Grafana dashboard ConfigMap is deployed. Defaults to the namespace of the HiveMQ Platform Operator.
```

charts/hivemq-platform-operator/values.yaml
```diff
@@ -164,3 +164,32 @@ crd:
...
+    # namespace: monitoring
+    # Overrides the default HiveMQ Platform Grafana dashboard by providing the name of a Kubernetes ConfigMap
```
mhofsche, line 178:
```suggestion
    # The name of a Kubernetes ConfigMap with a custom HiveMQ Platform Grafana dashboard to override the default dashboard.
```

charts/hivemq-platform-operator/values.yaml
```diff
@@ -164,3 +164,32 @@ crd:
...
+    # that contains the Grafana dashboard (also set create=false).
```
mhofsche, line 179:
```suggestion
    # (also set create=false).
```

charts/hivemq-platform-operator/values.yaml
```diff
@@ -164,3 +164,32 @@ crd:
...
+    # Custom Grafana dashboard ConfigMap content should be in JSON format along with the label `grafana_dashboard: "1"`
```
mhofsche, line 180:
```suggestion
    # Grafana dashboard ConfigMap content must be in JSON format along with the label `grafana_dashboard: "1"`
```

charts/hivemq-platform-operator/values.yaml
```diff
@@ -164,3 +164,32 @@ crd:
...
+    name: ""
+    # Overrides the default HiveMQ Platform Grafana Operator dashboard content via file using --set-file grafana-dashboard.json.
+    overrideDashboard: ""
+  serviceMonitor:
+    # Namespace where the ServiceMonitor will be deployed to. Defaults to use the same namespace where the HiveMQ Platform Operator Helm chart is deployed to.
```
mhofsche, line 185:
```suggestion
    # Namespace where to deploy the ServiceMonitor. Defaults to the namespace of the HiveMQ Platform Operator.
```

charts/hivemq-platform-operator/values.yaml
```diff
@@ -164,3 +164,32 @@ crd:
...
+    # namespace: monitoring
+    # Helm chart release name of the Prometheus Monitoring stack available in the Kubernetes cluster.
```
mhofsche, line 187:
```suggestion
    # Your Helm chart release name for the Prometheus Monitoring stack deployed in the Kubernetes cluster.
```

charts/hivemq-platform-operator/values.yaml
```diff
@@ -164,3 +164,32 @@ crd:
...
+    # This allows the Prometheus operator to discover the HiveMQ Platform Operator ServiceMonitor. Defaults to the HiveMQ Platform Operator Helm chart release name.
```
mhofsche, line 188:
```suggestion
    # The release name allows the Prometheus operator to discover the HiveMQ Platform Operator ServiceMonitor. Defaults to the HiveMQ Platform Operator Helm chart release name.
```

mhofsche (approved):
review of the values.yaml only

charts/hivemq-platform-operator/files/grafana-dashboard.json
Donnerbart: Fixed both issues in the operator repo, please use https://github.com/hivemq/hivemq-platform-operator/blob/develop/src/test/scripts/helm/prometheus/grafana-dashboard-operator.yml

charts/hivemq-platform-operator/values.yaml
```diff
@@ -164,3 +164,32 @@ crd:
...
+  # Dashboard configuration options.
+  dashboard:
+    # Creates a default HiveMQ Platform Operator Grafana dashboard ConfigMap.
+    create: true
+    # Namespace where to deploy the Grafana dashboard ConfigMap. Defaults to the namespace of the HiveMQ Platform Operator.
+    # namespace: monitoring
+    # The name of a Kubernetes ConfigMap with a custom HiveMQ Platform Operator Grafana dashboard to override the default dashboard.
+    # (also set create=false).
+    # Grafana dashboard ConfigMap content must be in JSON format along with the label `grafana_dashboard: "1"`
+    name: ""
```
Donnerbart, line 181: I don't understand the use-case of this config option. If we have `create=false`, why do we need to know an external ConfigMap name here? So far I only found one test that checks if the name is valid, but no further usage. What do we do with this when it's valid?

charts/hivemq-platform/tests/monitoring/hivemq_service_monitor_test.yaml
```diff
@@ -0,0 +1,202 @@
...
+        enabled: true
+        serviceMonitor:
+          interval: 1s
+          scrapeTimeout: 2s
+    asserts:
+      - contains:
+          path: spec.endpoints
+          content:
+            port: metrics-9499
+            path: /metrics
```
Donnerbart, line 158: Super nice that this works so flawless!

charts/hivemq-platform/tests/monitoring/hivemq_grafana_dashboard_test.yaml
```diff
@@ -0,0 +1,9631 @@
...
+  - it: with monitoring enabled and default values, ConfigMap spec created with defaults values
+    set:
+      monitoring.enabled: true
+    asserts:
+      - exists:
+          path: data["hivemq-dashboard.json"]
+      - equal:
+          path: data["hivemq-dashboard.json"]
+          value: |
+            {
```
Donnerbart, line 158: Uff :sweat_smile: Maybe we can persist a snapshot for this and just check that the value hasn't changed? https://github.com/helm-unittest/helm-unittest/blob/main/README.md#snapshot-testing

Then we wouldn't need to inline the whole JSON again and would only need to update the snapshow when the dashboard JSON is updated.

Donnerbart: We could also file a feature request to verify against a file, e.g.:
```
valueFrom:
  file: "files/grafana-dashboard.json"
```

charts/hivemq-platform-operator/values.yaml
afalhambra-hivemq: Very well spotted! Removing as not needed.

charts/hivemq-platform/tests/monitoring/hivemq_grafana_dashboard_test.yaml
afalhambra-hivemq: That's pretty awesome! Thanks for the suggestion, didn't know that one.
Will file a feature request as well.

Donnerbart: I think we need to check in these files as well, otherwise the assertion will do nothing when executed for the first time (e.g. on each GitHub test execution):
```
        charts/hivemq-platform-operator/tests/__snapshot__/
        charts/hivemq-platform/tests/monitoring/__snapshot__/
```

Donnerbart: And I think we have to be careful to not change the test name/description, since that seems to be the key in the snapshot files.

Donnerbart: It kind of makes sense. The assertion is "the value of `path: data["hivemq-dashboard.json"]` must match the snapshot of `it: with monitoring enabled and default values, ConfigMap spec created with defaults values`".

And the snapshot file matches the testsuite file name and contains snapshots for each test like:
```yaml
with monitoring enabled and default values, ConfigMap spec created with defaults values:
  1: |
    ...
```

Donnerbart: I've added the two files and also fixed the branch protection rules, so the PR can become green.

afalhambra-hivemq: https://github.com/helm-unittest/helm-unittest/issues/500

afalhambra-hivemq: Thanks! I forgot to include the `.snap` files

PR #425, opened 2025-01-10
Custom annotations and labels for all platform resources along with `stringData` in License Secret

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/28888/details/

_Implementation notes_: We cannot use an auto detect mechanism here to check whether the data passed in the value is actually Base64 encoded or not, due to the use of `overrideLicense` which is always in cleared text. Right now, if customer provides the `data` that one has to be Base64 encoded, but when using the `overrideLicense` that has to be in cleared text as per the license file itself.
So in an auto discovery mechanism, when using the `overrideLicense` value it will always be using the `data` field in the Secret no matter what, instead of the `stringData`

### DIFF
--- a/charts/hivemq-platform/templates/hivemq-license.yml
+++ b/charts/hivemq-platform/templates/hivemq-license.yml
@@ -4,6 +4,14 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ .Values.license.name | default (include "hivemq-platform.default-license-name" .) }}
+  {{- with .Values.license.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with .Values.license.labels }}
+  labels:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
 data:
   {{- if .Values.license.data }}
   license.lic: {{ .Values.license.data }}
--- a/charts/hivemq-platform/templates/hivemq-license.yml
+++ b/charts/hivemq-platform/templates/hivemq-license.yml
@@ -12,16 +12,22 @@ metadata:
   labels:
     {{- toYaml . | nindent 4 }}
   {{- end }}
-data:
+{{ ternary (println "data:") (println "stringData:") (.Values.license.isLicenseBase64Encoded) }}
   {{- if .Values.license.data }}
   license.lic: {{ .Values.license.data }}
   {{- else if .Values.license.overrideLicense }}
   license.lic: |-
-    {{ range (.Values.license.overrideLicense | b64enc) | toStrings }}
-    {{ . | nindent 4 }}
-    {{ end }}
+    {{- if .Values.license.isLicenseBase64Encoded -}}
+      {{- range .Values.license.overrideLicense | b64enc | toStrings -}}
+      {{ . | nindent 6 }}
+      {{ end }}
+    {{- else -}}
+      {{- range .Values.license.overrideLicense | toStrings -}}
+      {{ . | nindent 6 }}
+      {{ end }}
+    {{- end }}
   {{- end }}
-  {{- include "hivemq-platform.generate-licenses-content" (dict "licenses" .Values.license.additionalLicenses "licenseExtension" ".lic") }}
-  {{- include "hivemq-platform.generate-licenses-content" (dict "licenses" .Values.license.extensions "licenseExtension" ".elic") }}
-  {{- include "hivemq-platform.generate-licenses-content" (dict "licenses" .Values.license.dataHub "licenseExtension" ".plic") }}
+  {{- include "hivemq-platform.generate-licenses-content" (dict "licenses" .Values.license.additionalLicenses "licenseExtension" ".lic" "isLicenseBase64Encoded" .Values.license.isLicenseBase64Encoded) }}
+  {{- include "hivemq-platform.generate-licenses-content" (dict "licenses" .Values.license.extensions "licenseExtension" ".elic" "isLicenseBase64Encoded" .Values.license.isLicenseBase64Encoded) }}
+  {{- include "hivemq-platform.generate-licenses-content" (dict "licenses" .Values.license.dataHub "licenseExtension" ".plic" "isLicenseBase64Encoded" .Values.license.isLicenseBase64Encoded) }}
 {{- end }}
--- a/charts/hivemq-platform/values.yaml
+++ b/charts/hivemq-platform/values.yaml
@@ -101,6 +101,10 @@ license:
   # The name of the Kubernetes Secret with all HiveMQ licenses.
   # To create a new Kubernetes Secret with licenses, set create=true. Otherwise, false to reuse an existing Secret.
   name: ""
+  # Defines specific annotations to be applied to the Kubernetes Secret that contains the HiveMQ licenses.
+  annotations: {}
+  # Defines specific labels to be applied to the Kubernetes Secret that contains the HiveMQ licenses.
+  labels: {}
   # Inlines the main HiveMQ license as an encoded 64-bytes string.
   data: ""
   # Overrides the main HiveMQ license via file using --set-file license.lic.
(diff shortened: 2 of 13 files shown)


charts/hivemq-platform/templates/hivemq-license.yml
```diff
@@ -4,16 +4,30 @@ apiVersion: v1
...
-data:
+  {{- with .Values.license.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with .Values.license.labels }}
+  labels:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+{{ ternary (println "data:") (println "stringData:") (.Values.license.isDataBase64Encoded) }}
```
Donnerbart, line 15: Wouldn't it be easier if we only allow `data` or `stringData` and fail the template if both are set? Then we can maybe reduce this complexity and only apply `isDataBase64Encoded for `data` and `overrideLicense`.

And for when `stringData` is set then `isDataBase64Encoded` must either be false or is ignored?

afalhambra-hivemq: > Wouldn't it be easier if we only allow data or stringData and fail the template if both are set? Then we can maybe reduce this complexity and only apply isDataBase64Encoded for data and overrideLicense.

Not completely sure if I understand what you mean here. 
We are not allowing `data` or `stringData` to the customer values and only `data/overrideLicense/isDataBase64Encoded` here. 
So base on the `isDataBase64Encoded` value provided we just use either `data` or `stringData` here.

Donnerbart: Aaah, sorry, I have misread our `values.yaml` and the schema. I thought we added `stringData` as a new option in our `values.yaml`. But it's an option on the Secret... Okay, that makes much more sense now.

charts/hivemq-platform/values.yaml
```diff
@@ -101,7 +101,14 @@ license:
   # The name of the Kubernetes Secret with all HiveMQ licenses.
   # To create a new Kubernetes Secret with licenses, set create=true. Otherwise, false to reuse an existing Secret.
   name: ""
-  # Inlines the main HiveMQ license as an encoded 64-bytes string.
+  # Defines specific annotations to be applied to the Kubernetes Secret that contains the HiveMQ licenses.
+  annotations: {}
+  # Defines specific labels to be applied to the Kubernetes Secret that contains the HiveMQ licenses.
+  labels: {}
+  # Configures whether to use a Base64 encoded string (data) or clear content (stringData) for the Kubernetes Secret.
+  isDataBase64Encoded: true
```
mhofsche, line 109: Lets use "dataIsBase64Encoded" ...

afalhambra-hivemq: what about?

- base64EncodedData
- encodeDataToBase64
- dataRequiresBase64

Donnerbart: `isLicenseBase64Encoded`

otherwise I like `encodeDataToBase64` the most

afalhambra-hivemq: love it! Thanks @Donnerbart!

mhofsche: Agreed:  "isLicenseBase64Encoded" sounds good.

Donnerbart (approved):
Very nice solution for this very complex part of the templating!

charts/hivemq-platform/values.yaml
```diff
@@ -101,7 +101,24 @@ license:
   # The name of the Kubernetes Secret with all HiveMQ licenses.
   # To create a new Kubernetes Secret with licenses, set create=true. Otherwise, false to reuse an existing Secret.
   name: ""
-  # Inlines the main HiveMQ license as an encoded 64-bytes string.
+  # Defines specific annotations to be applied to the Kubernetes Secret that contains the HiveMQ licenses.
+  annotations: {}
+  # Defines specific labels to be applied to the Kubernetes Secret that contains the HiveMQ licenses.
+  labels: {}
+  # Configures whether to use a Base64 encoded string (data) or clear content (stringData) for the rendered Kubernetes Secret.
```
mhofsche, line 108: Indicates whether the license data is provided as Base64 encoded string (data) or clear content (stringData) for the rendered Kubernetes Secret.

charts/hivemq-platform/values.yaml
```diff
@@ -101,7 +101,24 @@ license:
...
+  # This is useful when using some third party tools to handle the license content through placeholders, such as the ArgoCD Vault plugin.
```
mhofsche, line 109: This is useful for third party tools that handle license content through placeholders, such as the ArgoCD Vault plugin.

charts/hivemq-platform/values.yaml
```diff
@@ -101,7 +101,24 @@ license:
...
+  # For example, the following Helm command:
+  #   `helm template platform hivemq/hivemq-platform --set license.create=true --set license.isLicenseBase64Encoded=false --set license.data=<hivemq-license-placeholder>`
+  # will render the following Kubernetes Secret:
```
mhofsche, line 112: renders the the following Kubernetes Secret:

charts/hivemq-platform/values.yaml
```diff
@@ -101,7 +101,24 @@ license:
...
+  #   apiVersion: v1
+  #   kind: Secret
+  #   metadata:
+  #     name: hivemq-license-platform
+  #   stringData:
+  #     license.lic: <hivemq-license-placeholder>
+  isLicenseBase64Encoded: true
+  # Inlines the main HiveMQ license as a string.
+  # By default, this data must be a Base64 encoded string. Otherwise, set isLicenseBase64Encoded=false to use a clear string.
```
mhofsche, line 121: By default, this data is expected as Base64 encoded string. Set "isLicenseBase64Encoded=false" to provide clear content.

charts/hivemq-platform/values.yaml
```diff
@@ -110,7 +127,8 @@ license:
   additionalLicenses: {}
     ## Defines the name of the additional main HiveMQ license (the .lic suffix will be added automatically).
     #license-name:
-    #  # Inlines the additional main HiveMQ license as an encoded 64-bytes string.
+    #  # Inlines the additional main HiveMQ license as a string.
+    #  # By default, this data must be a Base64 encoded string. Otherwise, set isLicenseBase64Encoded=false to use a clear string.
```
mhofsche, line 131: By default, this data is expected as Base64 encoded string. Set "isLicenseBase64Encoded=false" to provide a clear string.

charts/hivemq-platform/values.yaml
```diff
@@ -119,7 +137,8 @@ license:
   extensions: {}
     ## Defines the name of the HiveMQ Enterprise Extension license in the Kubernetes Secret (the .elic suffix will be added automatically).
     #extension-license-name:
-    #  # Inlines the HiveMQ Enterprise Extension license as an encoded 64-bytes string.
+    #  # Inlines the HiveMQ Enterprise Extension license as a string.
+    #  # By default, this data must be a Base64 encoded string. Otherwise, set isLicenseBase64Encoded=false to use a clear string.
```
mhofsche, line 141: By default, this data is expected as Base64 encoded string. Set "isLicenseBase64Encoded=false" to provide a plain string.

charts/hivemq-platform/values.yaml
```diff
@@ -134,7 +153,8 @@ license:
   dataHub: {}
     ## Defines the name of the HiveMQ Data Hub license in the Kubernetes Secret (the .plic suffix will be added automatically).
     #data-hub-license-name:
-    #  # Inlines the HiveMQ Data Hub license as an encoded 64-bytes string.
+    #  # Inlines the HiveMQ Data Hub license as a string.
+    #  # By default, this data must be a Base64 encoded string. Otherwise, set isLicenseBase64Encoded=false to use a clear string.
```
mhofsche, line 157: By default, this data is expected as Base64 encoded string. Set "isLicenseBase64Encoded=false" to provide a plain string.

PR #440, opened 2025-01-22
Add matrix test for minimum and latest K8s version

Donnerbart:
https://hivemq.kanbanize.com/ctrl_board/22/cards/29641/details/

### DIFF
--- a/tests-hivemq-operator/build.gradle.kts
+++ b/tests-hivemq-operator/build.gradle.kts
@@ -1,3 +1,5 @@
+import com.fasterxml.jackson.dataformat.toml.TomlMapper
+
 plugins {
     java
 }
@@ -14,6 +16,15 @@ java {
 
 repositories {
     mavenCentral()
+}
+
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath(libs.jackson.dataformat.toml)
+    }
 }
 
 @Suppress("UnstableApiUsage")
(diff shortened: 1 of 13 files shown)


tests-hivemq-operator/build.gradle.kts
```diff
@@ -65,7 +76,25 @@ testing {
...
+                        // sets Docker image tags for the tests
+                        val tomlFile = projectDir.resolve("gradle").resolve("docker.versions.toml")
+                        val tomlDocker = TomlMapper().readTree(tomlFile).path("docker")
+                        tomlDocker.fields().forEach { (key, value) ->
+                            val tag = value.path("tag").asText()
+                            if (tag.isNotEmpty()) {
+                                println("Configuring test Docker image $key:$tag")
+                                systemProperty("$key.tag", tag)
+                            }
+                        }
```
afalhambra-hivemq, line 89: I believe this loop is not really needed, as the `k3s.tag` is the only one used in the legacy operator ITs. And no more custom images are coming either from the `docker.version.toml` file.
Though it may be good in case we added some other image to that file and use it afterwards in an IT... probably not.
```suggestion
```

tests-hivemq-operator/build.gradle.kts
```diff
@@ -65,7 +76,25 @@ testing {
...
+                        systemProperty("hivemq.tag", libs.versions.hivemq.platform.get())
+                        val k8sVersionType = environment["K8S_VERSION_TYPE"] ?: "LATEST"
+                        val key = if (k8sVersionType == "MINIMUM") "k3s-minimum" else "k3s-latest"
+                        val tag = tomlDocker.path(key).path("tag").asText()
+                        println("Configuring test Docker image k3s:$tag ($k8sVersionType)")
+                        systemProperty("k3s.tag", tag)
+                        systemProperty("k3s.version.type", k8sVersionType)
```
afalhambra-hivemq, line 96: There is no `k3s.version.type` used in the legacy operator ITs, if I'm not wrong.
```suggestion
```

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/compose/HelmLegacyStatefulSetMigrationIT.java
```diff
@@ -79,6 +80,9 @@ void tearDown() throws Exception {
 
     @Test
     @Timeout(value = 5, unit = TimeUnit.MINUTES)
+    @EnabledIfSystemProperty(named = "k3s.version.type",
```
afalhambra-hivemq, line 83: Nice!

afalhambra-hivemq (approved):
Great work adding this new matrix for the K8s versions!

tests-hivemq-operator/build.gradle.kts
Donnerbart: Yeah, I would just like to keep both scripts in sync. Just in case we would need to add nginx here as well (let's hope that we don't have to).

tests-hivemq-operator/build.gradle.kts
Donnerbart: It's not used, but same, would like to keep both in sync.

PR #444, opened 2025-01-23
Create Roles and RoleBindings instead of ClusterRoles and ClusterRoleBindings for the HiveMQ Platform Operator if namespaces are configured

Donnerbart:
https://hivemq.kanbanize.com/ctrl_board/22/cards/27640/details/

### DIFF
--- a/tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/platform/UpdateLogLevelIT.java
+++ b/tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/platform/UpdateLogLevelIT.java
@@ -1,0 +1,108 @@
+package com.hivemq.helmcharts.platform;
+
+import com.hivemq.helmcharts.AbstractHelmChartIT;
+import com.hivemq.helmcharts.util.K8sUtil;
+import io.fabric8.kubernetes.api.model.Pod;
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Timeout;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.concurrent.TimeUnit;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+
+@Tag("Platform")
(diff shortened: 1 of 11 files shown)


charts/hivemq-platform-operator/templates/_helpers_rbac.tpl
```diff
@@ -0,0 +1,145 @@
...
+      - delete
+  - apiGroups:
+      - ""
+    resources:
+      - pods/exec
+    verbs:
+      - create
+      - watch
+      - get
+  - apiGroups:
```
Donnerbart, line 101: TODO: remove

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/platform/UpdateLogLevelIT.java
```diff
@@ -0,0 +1,105 @@
...
+    private @NotNull Path logbackFile;
+
+    @Override
+    protected boolean installPlatformOperatorChart() {
+        return false;
+    }
+
+    @BeforeEach
+    @Timeout(value = 3, unit = TimeUnit.MINUTES)
+    void setUp(@TempDir final @NotNull Path tempDir) {
```
afalhambra-hivemq, line 38: Can't remember now if there was an issue with this way of using `@TempDir` if the temp folder was not properly cleanup afterwards? in different ITs we have it defined outside of the `setUp` method as a private class member.

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/platform/UpdateLogLevelIT.java
```diff
@@ -0,0 +1,105 @@
...
+            Files.deleteIfExists(logbackFile);
+            client.pods()
+                    .inNamespace(platformNamespace)
+                    .withName(pod.getMetadata().getName())
+                    .file("/opt/hivemq/conf/logback.xml")
+                    .copy(logbackFile);
+            final var logbackXml = Files.readString(logbackFile);
+            assertThat(logbackXml).contains(value);
+        } catch (final IOException e) {
+            throw new AssertionError("Could not copy logback.xml file from pod", e);
```
afalhambra-hivemq, line 102: Could not copy or delete `logback.xml`.

afalhambra-hivemq (approved):
Great refinement of the RBAC permissions!

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/platform/UpdateLogLevelIT.java
Donnerbart: I mostly do this as well, but looked into other helm tests and found this usage in `HelmPlatformMutualTlsIT`. But it's the only test that does it like this. Will align both.

PR #460, opened 2025-01-31
Update index.md

sfrehse:
fix(): Updates the index


afalhambra-hivemq:
@sfrehse I think you're targeting/using the wrong branch. This `gh-pages` are automatically managed by the Helm chart releaser.
If you need to make any modifications, you need to use the `master` branch for the same and fix the issue there. Otherwise, with a next release from `master` any changes into this branch may get overwritten.

afalhambra-hivemq (approved):

PR #462, opened 2025-01-31
Fix race condition in release notes updates and do not check previous appVersion for simple release note generation (used by hivemq-edge)

Donnerbart:
https://hivemq.kanbanize.com/ctrl_board/22/cards/29898/details/

We need to fix this, otherwise the release notes will not be updated if just the Chart is released for `hivemq-edge` without bumping the `appVersion` as well.

This also fixes the race condition between updating the `index.yaml` by the chart releaser, which triggers a deployment of the GitHub page that is used for our Helm repository, and using the Helm repository information to update the GitHub release notes. The solution is to download the updated `index.yaml` file instead of using the `helm repo update` (that needs the GitHub pages deployment to be done).

### DIFF
--- a/github-release-note-updater/src/main/java/com/hivemq/release/GitHubReleaseNotesUpdater.java
+++ b/github-release-note-updater/src/main/java/com/hivemq/release/GitHubReleaseNotesUpdater.java
@@ -2,6 +2,7 @@ package com.hivemq.release;
 
 import com.beust.jcommander.JCommander;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.github.zafarkhaja.semver.Version;
 import org.jetbrains.annotations.NotNull;
 
@@ -58,7 +59,7 @@ public class GitHubReleaseNotesUpdater {
             System.exit(0);
         }
 
-        final var chartsPath = Path.of(arguments.path, "charts.json");
+        final var chartsPath = Path.of(arguments.path, "charts.yaml");
         final var releasesPath = Path.of(arguments.path, "releases.json");
         if (Files.notExists(chartsPath)) {
             System.err.printf("Charts path '%s' does not exist%n", chartsPath);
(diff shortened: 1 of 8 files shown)


codepitbull (approved):
LGTM

github-release-note-updater/src/main/java/com/hivemq/release/IndexFile.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.Map;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+record IndexFile(String apiVersion, Map<String, List<Chart>> entries, String generated) {
+
+    @JsonCreator
+    IndexFile(
+            @JsonProperty("apiVersion") final String apiVersion,
+            @JsonProperty("entries") final Map<String, List<Chart>> entries,
+            @JsonProperty("generated") final String generated) {
```
afalhambra-hivemq, line 17:
```suggestion
            @JsonProperty("apiVersion") final @NotNull String apiVersion,
            @JsonProperty("entries") final @NotNull Map<String, List<Chart>> entries,
            @JsonProperty("generated") final @NotNull String generated) {
```

github-release-note-updater/src/main/java/com/hivemq/release/GitHubReleaseNotesUpdater.java
```diff
@@ -206,19 +208,13 @@ private static void setReleaseNotes(
             final @NotNull List<Chart> charts,
             final @NotNull String releaseNoteTemplate,
             final @NotNull Function<Version, String> releaseUrlFunction) {
-        for (int i = 0; i < charts.size(); i++) {
-            // we get the previous chart and check if the appVersion has changed in the current chart
-            final var chart = charts.get(i);
-            final var previousChart = i == 0 ? null : charts.get(i - 1);
-            final var wasChartUpdated = previousChart == null || !previousChart.appVersion().equals(chart.appVersion());
```
afalhambra-hivemq, line 213: So this would only be needed in case a previous chart was already set with a different release notes, right? But I'm wondering now whether we should've checked for the Chart version instead of the Chart appVersion in the first place (we could've released a new Chart only, without any app version changed, though probably this will not require a blog post I guess).

afalhambra-hivemq (approved):

github-release-note-updater/src/main/java/com/hivemq/release/GitHubReleaseNotesUpdater.java
Donnerbart: This method is only used for edge at the moment, so the restriction must be gone.

For the other non-edge charts I don't think the chart version will work. Because the chart version is always updated. But we need to know which app version update triggered the release (operator or platform). So I think we have to stick with the app version for the other charts.

PR #471, opened 2025-02-07
Release HiveMQ Edge 2025.2

codepitbull:


charts/hivemq-edge/tests/license/hivemq_edge_statefulset_test.yaml
```diff
@@ -54,7 +54,7 @@ tests:
           value: hivemq-edge
       - equal:
           path: spec.template.spec.containers[0].image
-          value: hivemq/hivemq-edge:2025.1-k8s
+          value: hivemq/hivemq-edge:2025.2
```
afalhambra-hivemq, line 57: regarding the failing Helm chart unit tests, you could set up the container image version in the unit test itself to avoid having to keep on changing the test itself with each and every release (can be also set at a test suite level if desired). 

```yaml
  - it: Should define the correct container properties
    set:
      image.repository: hivemq/hivemq-edge
      image.name: edge-tag
    asserts:
      - equal:
          path: spec.template.spec.containers[0].image
          value: hivemq/hivemq-edge:edge-tag
```

codepitbull: Fixed

PR #487, opened 2025-02-25
Update version to 4.37.0

ymengesha:


afalhambra-hivemq (approved):
Missing this one: https://github.com/hivemq/helm-charts/blob/update-version/charts/hivemq-edge/templates/tests/test-mqtt-cli.yml#L10

I will update the Gradle task to include this file, but in the meantime it's fine to update it manually

PR #494, opened 2025-03-06
Use Gradle Oci

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/30389/details/

### DIFF
--- a/tests-hivemq-platform-operator/build.gradle.kts
+++ b/tests-hivemq-platform-operator/build.gradle.kts
@@ -1,7 +1,19 @@
 import com.fasterxml.jackson.dataformat.toml.TomlMapper
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+import org.gradle.api.tasks.testing.logging.TestLogEvent
+
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath(libs.jackson.dataformat.toml)
+    }
+}
 
 plugins {
     java
+    alias(libs.plugins.oci)
 }
 
 group = "com.hivemq.helmcharts"
@@ -14,15 +26,6 @@ java {
 
 repositories {
     mavenCentral()
-}
-
-buildscript {
-    repositories {
-        mavenCentral()
-    }
-    dependencies {
-        classpath(libs.jackson.dataformat.toml)
-    }
 }
 
 @Suppress("UnstableApiUsage")
@@ -64,7 +67,14 @@ testing {
             targets.configureEach {
                 testTask {
                     testLogging {
-                        events("started", "passed", "skipped", "failed")
+                        events = setOf(
+                            TestLogEvent.STARTED,
+                            TestLogEvent.PASSED,
+                            TestLogEvent.SKIPPED,
+                            TestLogEvent.FAILED,
+                            TestLogEvent.STANDARD_ERROR,
+                        )
+                        exceptionFormat = TestExceptionFormat.FULL
                         showStandardStreams = true
                     }
                     reports {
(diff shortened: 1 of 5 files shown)


tests-hivemq-platform-operator/build.gradle.kts
```diff
@@ -95,8 +105,8 @@ testing {
                         layout.buildDirectory.file("hivemq-dns-init-wait.tar"),
                         layout.buildDirectory.file("hivemq-operator.tar"),
                         layout.buildDirectory.file("hivemq-k8s.tar"),
-                        layout.buildDirectory.file("hivemq-platform-operator-init.tar"),
-                        layout.buildDirectory.file("hivemq-platform-operator.tar"),
+                        layout.buildDirectory.file("test-image-tars/hivemq-platform-operator-init.tar"),
```
LukasBrand, line 108: My mistake:
Task `operatorImageLayoutForTestingTar` writes the tar file to the root build directory. Here it is required from the `test-image-tars` subdirectory.

I have no clue why this still works.

I also got it wrong inside the k8s-testsuite.

Easiest would be to just remove the folder here again.

tests-hivemq-platform-operator/build.gradle.kts
```diff
@@ -95,8 +105,8 @@ testing {
...
+                        layout.buildDirectory.file("test-image-tars/hivemq-platform-operator.tar"),
```
LukasBrand, line 109: Same applies here

tests-hivemq-platform-operator/build.gradle.kts
afalhambra-hivemq: mmm I guess it's because the tar files are still stored in under the root build directory. Below after a clean build:

image

Will change to root build directory

afalhambra-hivemq: Changed as well on the https://github.com/hivemq/hivemq-k8s-testsuite/pull/144/commits/d669e5ff915e2428d9a614bd1c829b9d4e647fb0

afalhambra-hivemq: Moved everything now to the `build/test-image-tars/` location instead.

tests-hivemq-platform-operator/build.gradle.kts
```diff
@@ -139,7 +160,13 @@ val savePlatformDockerImage by tasks.registering(Exec::class) {
...
     dependsOn(pullPlatformDockerImage)
-    workingDir(layout.buildDirectory)
+    val outputDir = layout.buildDirectory.dir("test-image-tars").get().asFile
+    doFirst {
+        if (!outputDir.exists()) {
+            outputDir.mkdirs()
+        }
+    }
+    workingDir(outputDir)
     commandLine("docker", "save", "-o", "hivemq-platform.tar", "docker.io/hivemq/hivemq4:$hivemqVersion")
```
LukasBrand, line 170: The task should define an output.
```
outputs.file(outputDir.file("hivemq-platform.tar"))
```
Syntax could be wrong, I just drafted it in my head...

LukasBrand (approved):

tests-hivemq-platform-operator/build.gradle.kts
```diff
@@ -92,12 +102,9 @@ testing {
                     }
                     dependsOn(saveDockerImages)
                     inputs.files(
-                        layout.buildDirectory.file("hivemq-dns-init-wait.tar"),
```
LukasBrand, line 95: I assume the deleted inputs are not required?

afalhambra-hivemq: No, they're not. In these ITs there is only one place we we use the legacy operator (along with the DNS and K8s deleted images), but it's pulled out from DockerHub and not built locally.
I guess this is OK for now cause it's only used in one single IT and with the legacy operator being sunsetted soon, I'd keep it as is.

PR #515, opened 2025-03-26
Add HiveMQ Cluster Failure Detection to HiveMQ Platform chart

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/31238/details/

### DIFF
--- a/charts/hivemq-platform/templates/_config-xml.tpl
+++ b/charts/hivemq-platform/templates/_config-xml.tpl
@@ -169,9 +169,52 @@ Usage: {{ include "hivemq-platform.default-hivemq-configuration" . }}
     <discovery>
       <extension/>
     </discovery>
+    {{- $clusterFailureDetectionConfig := .Values.hivemqClusterFailureDetection }}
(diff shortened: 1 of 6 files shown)


Donnerbart:
What happens if I only configure the `enabled` flag and omit the other values? For example:
```
      hivemqClusterFailureDetection:
        heartbeat:
          enabled: true
```
I cannot see any default values, so I assume this fails due to the schema validation? That might be a missing test case as well (also for the TCP failure detection).

mhofsche (approved):

Donnerbart:
Each key (`enabled`, `interval`, `timeout`, ...) is individually checked and only added if set. We don't set default values in the `config.xml` if something is missing, but the broker will do. So it should be totally fine to just set some of the values, not all of them.

These combination are just not tested. We have unit tests for all kind of invalid values or complete sets of parameters. But no tests for partial configurations like:
```
    set:
      hivemqClusterFailureDetection:
        heartbeat:
          enabled: false
```
```
    set:
      hivemqClusterFailureDetection:
        heartbeat:
          enabled: true
          interval: 50
```
```
    set:
      hivemqClusterFailureDetection:
        heartbeat:
          timeout: 10
```
Those are all valid configurations, aren't they? And it depends on the broker how the actual configuration will look like (e.g. if `enabled` is `true` or `false` by default).

Donnerbart (approved):

PR #536, opened 2025-04-25
We were labeling the colume claims with the helm-chart and container version which made them not updateable

codepitbull:
Volume claims need to be reused across releases.
Since we were using the chart version in the labels this caused them to be non-updateable.

This PR removes all dynamoc labels and only leaves the static ones in place for volume claims.


caoccao (approved):
LGTM

charts/hivemq-edge/tests/datahub/hivemq_edge_statefulset_test.yaml
```diff
@@ -173,6 +173,17 @@ tests:
             - mountPath: /license.edgelic
               name: license
 
+  - it: Should define the correct volumeClaimTemplates
```
afalhambra-hivemq, line 176:
```suggestion
  - it: Should define the correct volumeClaimTemplates metadata
```

afalhambra-hivemq (approved):
@codepitbull - In case you want a new Helm chart release, you need to also bump the Chart version. Otherwise, after merging, there won't be anything released.

PR #549, opened 2025-05-09
Add init containers support for HiveMQ Swarm Helm chart

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/31798/details/

### DIFF
--- a/charts/hivemq-swarm/values.yaml
+++ b/charts/hivemq-swarm/values.yaml
@@ -40,6 +40,8 @@ commander:
   ## Add environment variables to the commander
   ##
   env: []
+  ## Additional init containers for the HiveMQ Commander pod
+  additionalInitContainers: []
   ## Additional volumes to the Swarm pod.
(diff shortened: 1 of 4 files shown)


charts/hivemq-swarm/values.yaml
```diff
@@ -40,6 +40,8 @@ commander:
   ## Add environment variables to the commander
   ##
   env: []
+  ## Additional init containers for the HiveMQ Commander pod
+  initContainers: []
```
afalhambra-hivemq, line 44: Not sure whether use `initContainers` or `additionalInitContainers`. There is no InitContainer in Swarm by default, so I decided to use this name.
Though, additional volumes and volumeMounts are called here `extraVolumes` and `extraVolumeMounts`

Donnerbart: `extraInitContainers` sounds weird, but would probably be the most consistent naming.

Donnerbart (approved):
Nice and clean addition!

PR #552, opened 2025-05-13
Make WebSocket path configurable

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/32954/details/

### DIFF
--- a/charts/hivemq-platform/values.schema.json
+++ b/charts/hivemq-platform/values.schema.json
@@ -886,6 +886,11 @@
             "maxLength" : 63,
             "type" : "string"
           },
+          "path" : {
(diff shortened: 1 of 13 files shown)


charts/hivemq-platform/values.schema.json
```diff
@@ -886,6 +886,11 @@
             "maxLength" : 63,
             "type" : "string"
           },
+          "path" : {
```
afalhambra-hivemq, line 889: This can be hard. I decided to go for `path` but we may also use `hivemqWebsocketPath`, `websocketPath`. Not sure. At the moment this is only applicable to `websocket` services type.
`hivemqWebsocketPath` might be aligned with some HiveMQ configuration options, but since this is pretty related to the "service/listener" I chose `path` only.

Donnerbart (approved):

PR #589, opened 2025-06-26
Use OCI registry to get rid of Docker TAR files

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/22067/details/


tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/testcontainer/HelmChartContainer.java
```diff
@@ -495,12 +459,23 @@ public void upgradePlatformChart(final @NotNull String releaseName, final @NotNu
...
-        final var dockerfile = Path.of(MountableFile.forClasspathResource("helm.dockerfile").getFilesystemPath());
-        final var imageName = new ImageFromDockerfile().withDockerfile(dockerfile)
-                .withBuildArg("K3S_TAG", K3S_DOCKER_IMAGE.getVersionPart())
-                .get();
-        return DockerImageName.parse(imageName).asCompatibleSubstituteFor(K3S_DOCKER_IMAGE.getUnversionedPart());
+    @SuppressWarnings("HttpUrlsUsage")
+    private static @NotNull Path getRegistryPath() {
+        final var registry = URI.create("http://%s".formatted(K3S_DOCKER_IMAGE.getRegistry()));
+        final var registryContent = """
+                mirrors:
```
SgtSilvio, line 466: It should be possible to directly use the specific image name (`host.docker.internal:<port>/<imagename>`) for the pod containers instead of mirroring DockerHub.

afalhambra-hivemq: That's a great one, thanks!

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/util/NginxUtil.java
```diff
@@ -200,7 +199,7 @@ private static String getResourceAsBase64(final @NotNull String name) {
                 .withVolumes(volumes)
                 .addNewContainer()
                 .withName(NGINX_CONTAINER_NAME)
-                .withImage(DockerImageNames.NGINX_DOCKER_IMAGE.asCanonicalNameString())
+                .withImage("docker.io/nginx:latest")
```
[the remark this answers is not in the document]
SgtSilvio: if this happens on mac, did you configure host.docker.internal as an insecure registry? https://github.com/SgtSilvio/gradle-oci-junit-jupiter?tab=readme-ov-file#configure-docker-on-macos--windows

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/testcontainer/HelmChartContainer.java
SgtSilvio: `insecure_skip_verify` does not disable TLS, probably only skips certificate/hostname validation

afalhambra-hivemq: Really great finding! StackOverflow the old good friend! 
Am trying it out locally and works without any issues, but I guess it's because of the registry host in MacOS `host.docker.internal` being used and the additional setup done there for the K3s container `super.withExtraHost("host.docker.internal", "host-gateway")`.

However, on Linux, it's coming as `localhost` which then it means that the registry won't be reachable from inside the pod running in the K3s cluster:

```
err="failed to pull and unpack image \"localhost:34721/hivemq/hivemq-platform-operator:snapshot\": 
failed to resolve reference \"localhost:34721/hivemq/hivemq-platform-operator:snapshot\": 
failed to do request: Head \"https://localhost:34721/v2/hivemq/hivemq-platform-operator/manifests/snapshot\": 
dial tcp [::1]:34721: connect: connection refused" image="localhost:34721/hivemq/hivemq-platform-operator:snapshot"
```

I guess more further container network setup is needed here with this setup, but it may be trickier.

I guess that's why it works for both MacOS and Linux when setting the mirror for `docker.io` to the local registry within the K3s container itself.

SgtSilvio: Sorry, my suggestion was wrong.
Instead of
```java
final var registry = K3S_DOCKER_IMAGE.getRegistry();
```
you need to use
```java
final var registry = "host.docker.internal:" + K3S_DOCKER_IMAGE.getRegistry().getPort();
```

SgtSilvio: Explanation why `OciImages.getImageName(<image>)` returns:
- `localhost:<port>/<image>` on Linux because docker runs on the host => `docker pull` runs on the host, so `localhost` needs to be used to access the registry that runs on the host
- `host.docker.internal:<port>/<image>`on Mac/Windows because docker runs in a VM => `docker pull` runs inside the VM, so `host.docker.internal` needs to be used to access the registry that runs on the host.

k3s runs as a container, so every pull inside the k3s container needs to access the host to reach the registry. This is done via routing to the ip referenced by the `host-gateway` variable, which is bound to the `host.docker.internal` dns name via `withExtraHost("host.docker.internal", "host-gateway")`

tests-hivemq-platform-operator/build.gradle.kts
```diff
@@ -83,31 +88,30 @@ testing {
...
-                    inputs.files(
-                        layout.buildDirectory.file("test-image-tars/hivemq-platform-operator-init.tar"),
-                        layout.buildDirectory.file("test-image-tars/hivemq-platform-operator.tar"),
-                        layout.buildDirectory.file("test-image-tars/hivemq-platform.tar"),
-                    )
                 }
             }
+            oci.of(this) {
+                imageDependencies {
+                    runtime(project).name("hivemq/helm-charts").tag(k3sTag)
```
SgtSilvio, line 104:
```suggestion
                    runtime(project).name("hivemq/helm-charts").tag("latest")
```

tests-hivemq-platform-operator/build.gradle.kts
```diff
@@ -83,31 +88,30 @@ testing {
...
-                        systemProperty("hivemq.tag", libs.versions.hivemq.platform.get())
                         val k8sVersionType = environment["K8S_VERSION_TYPE"] ?: "LATEST"
                         val key = if (k8sVersionType == "MINIMUM") "k3s-minimum" else "k3s-latest"
+                        val tomlFile = projectDir.resolve("gradle").resolve("docker.versions.toml")
+                        val tomlDocker = TomlMapper().readTree(tomlFile).path("docker")
                         val tag = tomlDocker.path(key).path("tag").asText()
-                        println("Configuring test Docker image k3s:$tag ($k8sVersionType)")
+                        println("Configuring test OCI image k3s:$tag ($k8sVersionType)")
                         systemProperty("k3s.tag", tag)
                         systemProperty("k3s.version.type", k8sVersionType)
```
SgtSilvio, line 98:
```suggestion
```

SgtSilvio, line 98:
```suggestion
```

afalhambra-hivemq: Ops... sorry, cannot do that, forgot why I had to keep it.

It's basically needed because of this check here:
https://github.com/hivemq/helm-charts/blob/improvement/22067-switch-to-oci-images/tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/testcontainer/HelmChartContainer.java#L112-L115

So I need to know the version of the K3s the test is currently running, otherwise, I cannot do that with the `latest` tag unfortunately.

afalhambra-hivemq: I'm gonna have a look to see if we can get rid of that check:
```java
                new ComparableVersion(K3S_DOCKER_IMAGE.getVersionPart()).compareTo(new ComparableVersion("1.24")) > 0 ?
                        "--disable=traefik" :
                        "--no-deploy=traefik",
```
This `traefik` is just an Ingress controller, so hopefully there is no impact in those tests using Nginx as a different Ingress controller.

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/util/NginxUtil.java
```diff
@@ -200,7 +200,7 @@ private static String getResourceAsBase64(final @NotNull String name) {
                 .withVolumes(volumes)
                 .addNewContainer()
                 .withName(NGINX_CONTAINER_NAME)
-                .withImage(DockerImageNames.NGINX_DOCKER_IMAGE.asCanonicalNameString())
+                .withImage(OciImages.getImageName("library/nginx").asCanonicalNameString())
```
afalhambra-hivemq, line 203:
```java
                .withImage(OciImages.getImageName("nginx").asCanonicalNameString())
```
does not work, the image cannot be found. Prefix `library` is needed as defined in the `build.gradle.kts` file.

SgtSilvio: yes, omitting `library` is just a special case when using the default docker hub registry

SgtSilvio: yes, you always need to use the dns name you specify via `withExtraHost("host.docker.internal", "host-gateway")` when pulling from inside the container. You still need to call `OciImages.getImageName(...)` and then change the hostname to get the correct port.

Donnerbart (approved):

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/util/OciUtil.java
```diff
@@ -0,0 +1,15 @@
+package com.hivemq.helmcharts.util;
+
+import io.github.sgtsilvio.gradle.oci.junit.jupiter.OciImages;
+import org.jetbrains.annotations.NotNull;
+
+public class OciUtil {
+
+    public static @NotNull String getImage(final @NotNull String imageName) {
```
SgtSilvio, line 8: Is this only used for the nginx image?
Why is this not used for all other images?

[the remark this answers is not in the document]
SgtSilvio: maybe move it as a static function to the `HelmChartContainer` class

afalhambra-hivemq: Yes, cause I cannot use it, neither here for the image ones loaded by Helm in the HelmChartContainer class:

```java
return Stream.of("--set", "image.repository=host.docker.internal/hivemq", "--set", "image.tag=snapshot");
...
return Stream.of("--set", "image.repository=host.docker.internal/hivemq", "--set", "image.tag=latest");
```

nor in the other images which are just only used _outside_ of the K3s cluster (they need just normal/regular OCI plugin used by using `OciImages.getImageName(...)`. Such as:

```java
    @Container
    private final @NotNull HiveMQContainer hivemqContainer =
            new HiveMQContainer(OciImages.getImageName("hivemq/hivemq4")) //
                    .withNetwork(network) //
                    .withNetworkAliases("remote") //
                    .withLogLevel(Level.DEBUG);
```

tests-hivemq-platform-operator/build.gradle.kts
afalhambra-hivemq: I just also added what @SgtSilvio  suggested:

> Maybe the --disable-default-registry-endpoint could be used (not sure how complicated it is to set it)
> 
> Nodes may be started with the --disable-default-registry-endpoint option. When this is set, containerd will not fall back to the default registry endpoint, and will only pull from configured mirror endpoints

So now we make sure we don't fallback into any DockerHub call (at least for latest versions of K3s, not for the minimum 1.23 though.

```java
  Objects.equals(System.getenv("K8S_VERSION_TYPE"), "LATEST") ?
          "--disable-default-registry-endpoint" :
          "",
```

PR #592, opened 2025-06-30
Update GitHub workflows

Donnerbart:


.github/workflows/hivemq-platform-operator-integration-test.yml
```diff
@@ -181,7 +181,7 @@ jobs:
         run: ./gradlew :tests-hivemq-platform-operator:integrationTest ${{ steps.split-tests.outputs.test-suite }}
 
       - name: Upload HiveMQ Platform Operator test results
-        if: always()
+        if: ${{ !cancelled() }}
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
```
afalhambra-hivemq, line 185: Should we then add the same condition/check to the next step as well `Upload JUnit report artifact`?

afalhambra-hivemq (approved):

.github/workflows/hivemq-platform-operator-integration-test.yml
Donnerbart: Discussed on Slack that we keep it for now (like in the operator)

PR #611, opened 2025-08-04
Improve RBAC permissions for HiveMQ Platform Operator

Donnerbart:
https://hivemq.kanbanize.com/ctrl_board/22/cards/34744/details/


charts/hivemq-platform-operator/templates/rbac.yml
```diff
@@ -15,6 +16,7 @@ rules:
   {{- $namespaceRules }}
```
afalhambra-hivemq, line 16: Am a little confused here now. For the ClusterRole rules, if the namespace is not set or empty (all namespaces), it should apply the namespace rules, if any namespace is set with one or more namespaces these rules are skipped.

PR #620, opened 2025-08-14
Add and update Platform Operator configuration options

afalhambra-hivemq:
https://hivemq.kanbanize.com/ctrl_board/22/cards/35535/details/

### DIFF
--- a/charts/hivemq-platform-operator/values.yaml
+++ b/charts/hivemq-platform-operator/values.yaml
@@ -128,13 +128,18 @@ podScheduling:
   #    operator: "Exists"
   #    effect: "NoSchedule"
 
-# TLS configuration for the Operator to access the Kubernetes API.
+# Configures the TLS options for the operator to access a secured Kubernetes API server.
 tls:
-  # Name of the Secret that contains keystore and truststore.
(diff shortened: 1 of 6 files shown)


charts/hivemq-platform-operator/values.yaml
```diff
@@ -18,9 +18,26 @@ resources:
...
+# Configures the Java Operator SDK log level for the HiveMQ Platform Operator.
+# Possible values are: "TRACE", "DEBUG", "INFO", "WARN", "ERROR"
+networkLogLevel: WARN
+
+# Configures the Java Operator SDK log level for the HiveMQ Platform Operator.
+# Possible values are: "TRACE", "DEBUG", "INFO", "WARN", "ERROR"
+sdkLogLevel: INFO
+
+# Configures whether to log the HiveMQ Platform Operator configuration at startup. Sensitive values are redacted.
+logConfiguration: false
```
Donnerbart, line 34: I would not expose these 3 options and just keep them in the `deployment.yml` template as they are, keeping the `.Values.xxx | default "..."` so they can be easily overridden in case we need to debug something.

Those are not needed for production and might just be confusing.

PR #650, opened 2025-09-15
Helm chart updates to expose UI pre-Login Message settings

marregui:
card: https://hivemq.kanbanize.com/ctrl_board/57/cards/35778/details/


afalhambra-hivemq:
@marregui not sure if you want to release the Edge charts with this PR? Asking cause if that's the case you need to bump the Chart version of the same.
Also, please change the target branch. `develop` branch is mainly used for the operator release. 
You should use `master` instead.

codepitbull (approved):

PR #856, opened 2026-04-08
Use unique Helm release names per test class

Donnerbart:
## Summary

- Fix flaky integration test failures caused by ClusterRole ownership collisions between test classes sharing the same K3s cluster
- Derive Helm release names from the test class name (e.g., `customlogback-op`) instead of using a shared static constant (`test-hivemq-platform-operator`), mirroring the existing per-test-class namespace pattern
- Add `getReleaseName()`, `getPlatformReleaseName()`, `getOperatorReleaseName()`, and `getLegacyReleaseName()` helpers to `K8sUtil`
- Add overridable `getPlatformReleaseName()` / `getOperatorReleaseName()` / `getLegacyReleaseName()` methods in `AbstractHelmChartIT` for tests whose truncated names would collide
- Add a `baseSetUp()` assertion that all release names stay within the `MAX_RELEASE_NAME_LENGTH` (33 chars) limit

### Problem

All integration tests used the same hardcoded operator release name `test-hivemq-platform-operator`. The operator Helm chart creates cluster-scoped resources (ClusterRole, ClusterRoleBinding) whose names include the release name. When a test's teardown timed out (e.g., `baseTearDown() timed out after 5 minutes`), the ClusterRole was left behind with a `meta.helm.sh/release-namespace` annotation pointing to the previous test's namespace. The next test in the same split would then fail with:

> `ClusterRole "hivemq-platform-operator-role-test-hivemq-platform-operator"` exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key `meta.helm.sh/release-namespace` must equal `<new-namespace>`: current value is `<old-namespace>`

This cascading failure was observed in [run #669](https://github.com/hivemq/helm-charts/actions/runs/24129688090) where `HelmCustomLogbackIT` teardown timed out and `HelmMqttCustomServiceNameIT` failed as a consequence.

### Solution

Each test class now gets a unique release name derived from its class name (with `Helm` prefix and `IT` suffix stripped, truncated to fit), with short suffixes for log identification:
- Platform: `<classname>-pf` (e.g., `customlogback-pf`)
- Operator: `<classname>-op` (e.g., `customlogback-op`)
- Legacy: `<classname>-lg` (e.g., `legacystatefulsetmigration-lg`)

Release names are capped at 33 characters (`K8sUtil.MAX_RELEASE_NAME_LENGTH`) to ensure all derived K8s resource names (the tightest being `hivemq-platform-<release>-dynamic-state`) stay within the 63-character K8s name limit.

Six test classes with long names that would collide after truncation override `getPlatformReleaseName()` / `getOperatorReleaseName()` with custom short names. A length assertion in `baseSetUp()` catches any future violations at test startup.

### DIFF
--- a/tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/migration/HelmLegacyStatefulSetMigrationIT.java
+++ b/tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/migration/HelmLegacyStatefulSetMigrationIT.java
@@ -45,14 +45,15 @@ import static org.assertj.core.api.Assertions.assertThat;
 class HelmLegacyStatefulSetMigrationIT extends AbstractHelmChartIT {
 
     private static final int MQTT_SERVICE_PORT = 1883;
-    private static final @NotNull String MQTT_SERVICE_NAME = "hivemq-%s-mqtt".formatted(LEGACY_RELEASE_NAME);
     private static final int CC_SERVICE_PORT = 8080;
-    private static final @NotNull String CC_SERVICE_NAME = "hivemq-%s-cc".formatted(LEGACY_RELEASE_NAME);
     private static final int METRICS_SERVICE_PORT = 9399;
-    private static final @NotNull String METRICS_SERVICE_NAME = "hivemq-%s-metrics".formatted(LEGACY_RELEASE_NAME);
     private static final int REST_API_SERVICE_PORT = 8888;
-    private static final @NotNull String REST_API_SERVICE_NAME = "hivemq-%s-api".formatted(LEGACY_RELEASE_NAME);
-    private static final @NotNull String CLUSTER_SERVICE_NAME = "hivemq-%s-cluster".formatted(LEGACY_RELEASE_NAME);
+
+    private final @NotNull String mqttServiceName = "hivemq-%s-mqtt".formatted(legacyReleaseName);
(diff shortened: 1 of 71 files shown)


tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/migration/HelmLegacyStatefulSetMigrationIT.java
```diff
@@ -210,63 +211,65 @@ void migrateToCurrentPlatformOperator() throws Exception {
         // ConfigMap by adding a new entry for `config.xml`.
         K8sUtil.updateConfigMap(client, operatorNamespace, "ese-legacy-config-map-update.yml");
 
-        helmChartContainer.installPlatformChart(LEGACY_RELEASE_NAME,
+        helmChartContainer.installPlatformChart(legacyReleaseName,
                 "-f",
                 "/files/migration-platform-stateful-set-values.yaml",
+                "--set",
+                "services[0].name=" + mqttServiceName,
```
afalhambra-hivemq: I think best would be to remove hardcoded MQTT service name from `migration-platform-stateful-set-values.yaml` and also the corresponding `--set services[0].name=` here.
The `legacyPortName` "mqtt" field already causes the chart to auto-generate the service name as `hivemq-<releaseName>-mqtt`, which matches the legacy operator's naming convention — making both the explicit name and the override redundant.

afalhambra-hivemq (approved):
Nice refactoring!

PR #868, opened 2026-04-14
Migrate HiveMQ Platform Operator tests to Java 25

Donnerbart:


tests-hivemq-platform-operator/build.gradle.kts
```diff
@@ -68,6 +68,12 @@ testing {
             }
             targets.configureEach {
                 testTask {
+                    jvmArgumentProviders.add(CommandLineArgumentProvider {
+                        listOf(
+                            "--enable-native-access=ALL-UNNAMED",
+                            "--sun-misc-unsafe-memory-access=allow",
```
afalhambra-hivemq, line 76: These 2 flags are needed because of Netty dependency mainly (included through hivemq-mqtt-client)? Native epoll access, and unsafe memory access?
Wondering whether it'd make sense to add a comment so we don't forget when to remove it

Donnerbart: `gradle-oci-junit-jupiter` and `hivemq-mqtt-client` have a Netty dependency.

The first warnings are actually coming from the OCI registry:
```
> Task :tests-hivemq-platform-operator:integrationTestSuiteOciRegistryData UP-TO-DATE
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::allocateMemory has been called by io.netty.util.internal.PlatformDependent0$2 (file:/home/donnerbart/.gradle/caches/modules-2/files-2.1/io.netty/netty-common/4.1.132.Final/14b8a487c8b9a105c718a44e3805cc538a08e523/netty-common-4.1.132.Final.jar)
WARNING: Please consider reporting this to the maintainers of class io.netty.util.internal.PlatformDependent0$2
WARNING: sun.misc.Unsafe::allocateMemory will be removed in a future release
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::loadLibrary has been called by io.netty.util.internal.NativeLibraryUtil in an unnamed module (file:/home/donnerbart/.gradle/caches/modules-2/files-2.1/io.netty/netty-common/4.1.132.Final/14b8a487c8b9a105c718a44e3805cc538a08e523/netty-common-4.1.132.Final.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled
```

Donnerbart: I think even with newer Netty 4.2 versions, some warnings will remain: https://netty.io/wiki/java-24-and-sun.misc.unsafe.html

Adding a comment to that blog article.

afalhambra-hivemq (approved):

PR #875, opened 2026-04-20
Fall back to base resource values for missing overrideLimits fields

afalhambra-hivemq:
## Summary

- When `nodes.resources.overrideLimits` is set with only a subset of fields (e.g., only `cpu`), the missing fields now fall back to base `nodes.resources` values instead of being omitted
- `overrideLimits: {}` now falls back to base resource values (limits = requests) instead of producing no limits at all
- Added `toString` to `memory` and `ephemeral-storage` pipelines to handle integer values safely

Resolves [INT-204](https://linear.app/hivemq/issue/INT-204) / [HIVEMQ-6095](https://hivemq.atlassian.net/browse/HIVEMQ-6095)


Donnerbart (approved):
Nice catch!

PR #894, opened 2026-05-06
Sign Harbor charts by tag and let cosign resolve the digest via the OCI registry endpoint

afalhambra-hivemq:
## Summary

Backport from `next-operator-release` (commit 38524a45).

The new repo-scoped Harbor robot account does not have `read-artifact` permission on Harbor's admin API (`/api/v2.0/...`), which the previous version of `sign-harbor-chart.sh` used to fetch the chart digest before signing. This caused signing to fail with HTTP 403.

This change drops the digest pre-fetch and lets `cosign sign` resolve the digest via the standard OCI registry endpoint (`/v2/.../manifests/...`), which only requires the `pull` permission already implied by the robot's push access.

The `concurrency: cancel-in-progress` rule on the workflow keeps the tag-vs-digest TOCTOU window closed in practice.

### DIFF
--- a/.github/workflows/harbor.yml
+++ b/.github/workflows/harbor.yml
@@ -6,6 +6,9 @@ on:
       - next-operator-release
     paths:
       - charts/hivemq-platform-operator/**
+concurrency:
+  group: publish-harbor-${{ github.ref }}
+  cancel-in-progress: true
 env:
   operator_chart_name: helm-charts/hivemq-platform-operator
 jobs:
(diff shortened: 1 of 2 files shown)


m-artin-white (commented):
The description references concurrency: cancel-in-progress, but that block exists on next-operator-release, not master, and this backport doesn't add it. Do you need backport the workflow change also?

afalhambra-hivemq:
> The description references concurrency: cancel-in-progress, but that block exists on next-operator-release, not master, and this backport doesn't add it. Do you need backport the workflow change also?

Great catch @m-artin-white - thank you!

m-artin-white (approved):

Donnerbart (approved):

PR #956, opened 2026-06-15
Assert retained messages by topic with received/missing summary

Donnerbart:
Subscribe to all topics on one client, drain the publish queue attributing each publish to its actual topic, and log received vs missing topics so a failure shows exactly which retained messages were not delivered.


afalhambra-hivemq (approved):
Nice small improvement!

PR #1010, opened 2026-08-02
Test the HiveMQ Platform on different operating systems and Java runtimes

Donnerbart:
## Summary

* Add a scheduled workflow that runs a subset of the HiveMQ Platform integration tests against HiveMQ Platform images built for the test run, one per operating system and Java runtime combination. The official image is built on Ubuntu with an Eclipse Temurin JRE, so nothing currently covers the runtimes customers actually use.
* Build the images with the gradle-oci plugin from public artifacts only: the platform distribution from `https://www.hivemq.com/releases/hivemq-latest.zip` on top of a public JRE base image. They are served to the K3s cluster of the test run and are never pushed to a registry.
* Add the `customPlatformImageVariant` Gradle property, which builds the image of that variant and replaces the official platform image in the test setup. Without the property, the build resolves the official image and runs the whole suite as before.
* Pin the JRE base images by digest in `oci.versions.toml`, so Renovate keeps them updated.
* Tag four existing integration tests with `custom-platform-image`, covering platform start-up, MQTT traffic, the extension system and two-node cluster forming with metrics. Only those run when the property is set. Tests that pin an image of their own stay untagged, since they would not exercise the variant.
* Generate the workflow matrix from the variant map in Gradle, so adding a variant means editing `oci.versions.toml` and that map, not the workflow.
* Document the variants and how to run them locally in the test module README.

## Variants

| Variant | Base image | JVM | Java | Covers |
| --- | --- | --- | --- | --- |
| `temurin21-resolute` | `eclipse-temurin:21-jre-resolute` | HotSpot | 21 | Java 21 on the newer base |
| `temurin21-ubi9` | `eclipse-temurin:21-jre-ubi9-minimal` | HotSpot | 21 | RHEL |
| `temurin25-ubi10` | `eclipse-temurin:25-jre-ubi10-minimal` | HotSpot | 25 | RHEL |
| `semeru21-noble` | `ibm-semeru-runtimes:open-21-jre-noble` | OpenJ9 | 21 | OpenJ9 |
| `semeru25-noble` | `ibm-semeru-runtimes:open-25-jre-noble` | OpenJ9 | 25 | OpenJ9 |
| `corretto25-al2023` | `amazoncorretto:25-al2023-jdk` | HotSpot | 25 | Amazon Linux |

The platform needs glibc, so musl-based images are out. Temurin publishes JRE images on `ubi9-minimal` up to Java 21 and on `ubi10-minimal` from Java 25, Semeru publishes JRE images for Ubuntu bases only, and Corretto ships no separate JRE for Java 11 and later. SUSE needs a layer that installs the distribution OpenJDK with `zypper` and is not part of this change.

## Notes

* The Java runtime base image is declared under its own module coordinate and routed to the real image with an `imageMapping` entry. The operator image builds on a Java runtime image of the same module, so declaring that module a second time makes Gradle's conflict resolution pick one version for the whole graph, which puts the operator on the Java runtime of the variant under test.
* Permissions are set explicitly on the image layer. Entries in the distribution zip are owned by root with `755`, while the container runs as uid 10000 and needs write access to `data`, `log`, `audit`, `backup` and `extensions`.
* `tools/` is excluded from the image, roughly 75 MB that no test uses.
* No entry point is configured, since the operator sets the container command itself so that the init application runs before the broker.
* The distribution is unpinned, so every run tests the current release. The tested version is written to the job summary, which keeps a failure attributable to a release.


.github/workflows/custom-platform-image-tests.yml
```diff
@@ -6,6 +6,10 @@
 # It shows that HiveMQ runs on a combination, nothing more, and is not a certification of it.
 name: HiveMQ Platform Custom Image Tests
 on:
+  # TODO: remove this trigger before merging, it only exists to run the workflow from the feature branch
+  push:
+    branches:
+      - feature/int-451-helm-charts-test-the-hivemq-platform-on-different-operating
```
afalhambra-hivemq: Just a reminder to remove this change

.github/workflows/custom-platform-image-tests.yml
```diff
@@ -0,0 +1,168 @@
...
+
+      - name: Generate variant list
+        id: generate
+        env:
+          VARIANT: ${{ inputs.variant }}
+        run: |
+          if [[ -n "$VARIANT" ]]; then
+            variants="[\"$VARIANT\"]"
+          else
+            variants=$(./gradlew :tests-hivemq-platform-operator:printCustomPlatformImageVariants -q --console=plain | tail -1)
```
afalhambra-hivemq: maybe an additional grep command in case a Gradle output displayed any other additional output (e.g. Gradle deprecated warnings).

```suggestion
            variants=$(./gradlew :tests-hivemq-platform-operator:printCustomPlatformImageVariants -q --console=plain | grep -oE '^\[.*\]$' | tail -1)
```

tests-hivemq-platform-operator/gradle/oci.versions.toml
```diff
@@ -45,3 +45,48 @@ pinnedReference = "4.7.10@sha256:241d6a8e1963968ebafb7f63396c56e1d8e41fc4065ee62
 name = "init-dns-wait"
 image = "hivemq/init-dns-wait"
 pinnedReference = "1.0.1@sha256:96f972c3029a2bab716c43c5404280030ecd98d199bd9cbe5de0f495a8f06945"
+
+# Java runtime base images for the custom HiveMQ Platform images, see the `platformImageVariant` property.
```
afalhambra-hivemq:
```suggestion
# Java runtime base images for the custom HiveMQ Platform images, see the `customPlatformImageVariant` property.
```

.github/workflows/custom-platform-image-tests.yml
```diff
@@ -0,0 +1,168 @@
...
+          ORG_GRADLE_PROJECT_dockerHubUsername: ${{ secrets.DOCKER_USERNAME }}
+          ORG_GRADLE_PROJECT_dockerHubPassword: ${{ secrets.DOCKER_TOKEN }}
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_TOKEN }}
+          K8S_VERSION_TYPE: LATEST
+          DEVELOCITY_SERVER_URL: ${{ secrets.DEVELOCITY_SERVER_URL }}
+          DEVELOCITY_ACCESS_KEY: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
+        run: >
+          ./gradlew :tests-hivemq-platform-operator:integrationTest
+          -PcustomPlatformImageVariant=${{ matrix.variant }}
```
afalhambra-hivemq:
```suggestion
          ./gradlew :tests-hivemq-platform-operator:integrationTest -PcustomPlatformImageVariant=${{ matrix.variant }}
```

tests-hivemq-platform-operator/build.gradle.kts
```diff
@@ -1,8 +1,11 @@
+import de.undercouch.gradle.tasks.download.Download
+import org.gradle.api.tasks.testing.junitplatform.JUnitPlatformOptions
```
afalhambra-hivemq: Seems that this import is redundant

.github/workflows/custom-platform-image-tests.yml
```diff
@@ -0,0 +1,168 @@
...
+      - feature/int-451-helm-charts-test-the-hivemq-platform-on-different-operating
+  schedule:
+    # every Sunday at 03:00 UTC
+    - cron: '0 3 * * 0'
+  workflow_dispatch:
+    inputs:
+      variant:
+        description: 'Custom platform image variant to test, all variants if empty'
+        required: false
+        type: string
```
afalhambra-hivemq, line 17: would it make sense to make it a `choice` instead? if the variant is invalid image, the Gradle task will fail down the line

```kotlin
  val jreBaseImage = jreBaseImages[customPlatformImageVariant]                                                                                                                                             
      ?: throw GradleException(                                                                                                                                                                            
          "Unknown custom platform image variant '$customPlatformImageVariant', " +                                                                                                                        
          "expected one of ${jreBaseImages.keys.sorted()}") 
```

Although this would introduce another place to keep in sync when the `jreBaseImages` property changed (not sure how frequent we will change an image though)

.github/workflows/custom-platform-image-tests.yml
```diff
@@ -0,0 +1,168 @@
+#file: noinspection UndefinedAction,UndefinedParamsPresent
```
afalhambra-hivemq:
```suggestion
#file: noinspection UndefinedAction,UndefinedParamsPresent,GithubFunctionSignatureValidation
```

above removes IntelliJ warning for the `fromjson` function call, line 72.
```
        variant: ${{ fromjson(needs.generate-variants-json.outputs.json) }}
```

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/extensions/HelmDataHubModuleIT.java
```diff
@@ -0,0 +1,181 @@
...
+ * the HiveMQ Platform stops, so the test also asserts that a restart produces no dynamic linker error.
+ */
+@Tag("custom-platform-image")
+class HelmDataHubModuleIT extends AbstractHelmChartIT {
+
+    private static final int MQTT_SERVICE_PORT = 1883;
+    private static final int REST_API_SERVICE_PORT = 8890;
+
+    // https://github.com/hivemq/hivemq-hello-world-datahub-module
+    private static final @NotNull String[] MODULE_FILES = {
```
afalhambra-hivemq:
```suggestion
    private static final @NotNull String @NotNull [] MODULE_FILES = {
```

tests-hivemq-platform-operator/src/integrationTest/java/com/hivemq/helmcharts/extensions/HelmDataHubModuleIT.java
```diff
@@ -0,0 +1,181 @@
...
+        pods.withName(podName)
+                .waitUntilCondition(pod -> pod != null &&
+                                !podUid.equals(pod.getMetadata().getUid()) &&
+                                !pod.getStatus().getContainerStatuses().isEmpty() &&
+                                pod.getStatus()
+                                        .getContainerStatuses()
+                                        .stream()
+                                        .allMatch(containerStatus -> Boolean.TRUE.equals(containerStatus.getReady())),
+                        3,
+                        TimeUnit.MINUTES);
```
afalhambra-hivemq: maybe we could improve or overload the `K8sUtil.waitForPodStateRunning` method and pass the pod UID as well, and maybe avoid passing any label? don't know, just trying to see whether it would make sense to try to unify and reuse that method.

.github/workflows/custom-platform-image-tests.yml
Donnerbart: Hmm, no idea, not sure how often we would fire an individual image. You can leave it empty, then it will just run all images.

.github/workflows/custom-platform-image-tests.yml
Donnerbart: Good one, that `tail -1` is fragile and could break with any Gradle update.

.github/workflows/custom-platform-image-tests.yml
Donnerbart: I will improve the validation, so we have a fail fast if you submit an unknown image.

PR #1023, opened 2026-08-07
[LATEST] Publish a prebuilt Helm image to ghcr and use it in the integration tests

Donnerbart:
## Summary

* Add a `helm-image` build that publishes `ghcr.io/hivemq/helm-test-image`, a public multi-arch image containing only the Helm binary at `/usr/local/bin/helm`.
* Build the image by downloading the Helm distribution for `linux/amd64` and `linux/arm64/v8` from `get.helm.sh` and verifying the published SHA-256 checksum. No container is started and no package manager runs, at test time or at publish time.
* Take the Helm version from `.helm-version`, so the integration tests stop running an arbitrary Helm 4 release and use the same pin as the workflows and the manifest scripts.
* Add a workflow that publishes on a change to `.helm-version` or to `helm-image/`, and on manual dispatch.
* Consume the prebuilt image in the operator and edge test builds as a parent image dependency next to the k3s image, replacing the per-run Helm layer build.
* Remove the `apt-get` step that failed intermittently on unhealthy Ubuntu mirrors, along with the Ubuntu base image and the `library/ubuntu` catalog entries. Nothing used the packages it contributed; the tests execute `helm` and `kubectl` only, and `kubectl` ships in the k3s image.

The image carries an OCI description marking it as test tooling rather than a supported artifact, and `helm-image/README.md` states the same.

The matching change for the Kubernetes test suite is in hivemq/hivemq-k8s-testsuite.

### DIFF
--- a/.github/workflows/helm-image.yml
+++ b/.github/workflows/helm-image.yml
@@ -1,0 +1,40 @@
+name: Publish Helm test image to ghcr
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - .helm-version
+      - helm-image/**
+  workflow_dispatch:
+concurrency:
+  group: publish-helm-image-${{ github.ref }}
+  cancel-in-progress: false
+jobs:
+  helm-image:
+    name: Helm test image
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout HiveMQ Helm Charts
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Set up JDK 25
+        uses: actions/setup-java@b6effb05e454b25005698d916606bdc6ffcbf961 # v5
+        with:
+          cache: gradle
+          distribution: temurin
+          java-version: 25
+
+      - name: Set up Gradle
+        uses: gradle/actions/setup-gradle@9c971963bec38e04b3d30dcc455b5382be2fdbfb # v6
+        with:
+          cache-disabled: true
+
+      - name: Publish Helm test image
+        env:
+          ORG_GRADLE_PROJECT_ghcrUsername: ${{ github.actor }}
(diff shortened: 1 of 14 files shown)


.github/workflows/helm-image.yml
```diff
@@ -0,0 +1,40 @@
+name: Publish Helm test image to ghcr
+on:
+  push:
```
afalhambra-hivemq, line 3: Should we wire this up with the dispatcher workflow, so that on a `pull_request` event any change to the new `helm-image` module or to the `.helm-version` file has to build successfully (e.g. `./gradlew :helm-image:ociImageLayout`, without pushing) before the PR can merge?
Otherwise this workflow only runs after the PR is merged, and if we screwed up anything there we'd only find out on master.

helm-image/settings.gradle.kts
```diff
@@ -0,0 +1,14 @@
...
+
+rootProject.name = "helm-image"
+
+develocity {
+    server = System.getenv("DEVELOCITY_SERVER_URL")
+    buildScan {
+        publishing.onlyIf { System.getenv("CI_RUN") == "true" }
+        uploadInBackground = false
+    }
+}
```
afalhambra-hivemq, line 14: Do we really need to setup Develocity here? I think we're not publishing any build scan to it in this PR

.github/workflows/helm-image.yml
Donnerbart: I'm adding a proper integration test and wire it up in the dispatcher, to close this gap. Good call!

helm-image/settings.gradle.kts
Donnerbart: The new test will be configured correctly with Tailscale and `CI_RUN`, so we publish a build scan.

afalhambra-hivemq (approved):
Nice improvement, we finally get rid of this apt get command!

PR #1030, opened 2026-08-12
[LATEST] Resolve the Helm and platform distributions as dependencies

Donnerbart:
## Summary

- `helm-image` fetched the Helm distribution and its `.sha256sum` with `de.undercouch.download` tasks, and `tests-hivemq-platform-operator` fetched the HiveMQ Platform distribution the same way. `Download` is not a cacheable task type and the destinations live in `build/`, so a clean build downloaded them again and no cache could carry them.
- Both are resolved through ivy repositories with a pattern layout now, over `get.helm.sh` and `www.hivemq.com/releases`. The artifacts land in `~/.gradle/caches/modules-2`, which persists on Jenkins agents and is restored across GitHub Actions jobs.
- The platform distribution moves from `hivemq-latest.zip` to the `hivemq-platform` version of the version catalog, which is the version the tests already use for `hivemq.tag` and the image tags. `updatePlatformVersion` rewrites that entry on every platform release, so the custom platform image keeps tracking releases while the build stays reproducible.
- The manual SHA-256 verification of the Helm archive is unchanged.
- With the last `Download` tasks gone, `de.undercouch.download` is removed from both modules and both version catalogs.

Verified by building `:helm-image:ociImageLayout` and the custom platform image layer: the Helm binary extracts and verifies, the platform layer holds the same 317 entries with the tools excluded and the distribution directory renamed, and a second run is up to date offline.

### DIFF
--- a/.github/workflows/custom-platform-image-tests.yml
+++ b/.github/workflows/custom-platform-image-tests.yml
@@ -19,6 +19,11 @@ concurrency:
   group: custom-platform-image-tests
 env:
   CI_RUN: true
+  # the same set of files in every job, so the warmup job and the matrix jobs share one Gradle cache entry
+  cache-dependency-path: |
+    helm-charts/**/*.gradle*
+    helm-charts/**/gradle-wrapper.properties
+    helm-charts/**/*.versions.toml
 permissions:
   contents: read
   checks: write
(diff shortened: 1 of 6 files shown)


afalhambra-hivemq (commented):
A couple of issues:

- This step needs to be update inside the `custom-platform-image-tests.yml` https://github.com/hivemq/helm-charts/blob/master/.github/workflows/custom-platform-image-tests.yml#L120
It is using the `build` folder used by the `download` plugin. Maybe just a `grep` command on the lib catalog can do the trick to fetch the HiveMQ version

- The tests-hivemq-platform-operator/README.md is outdated [here](https://github.com/hivemq/helm-charts/blob/master/tests-hivemq-platform-operator/README.md?plain=1#L56): 
```
The image is built from the latest platform distribution (`https://www.hivemq.com/releases/hivemq-latest.zip`)
```
Maybe something like
```
The image is built from the platform distribution of the hivemq-platform version in gradle/libs.versions.toml
```

afalhambra-hivemq (approved):

PR #1054, opened 2026-08-31
Platform Release 4.55.0

Donnerbart:
* Adds Data Intelligence clustering.
* Adds a simple persistence configuration to configure PVCs.


charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -94,24 +94,287 @@ Usage: {{ include "hivemq-platform.has-license" . }}
...
+        {{- fail (printf "\nThe release name `%s` exceeds 48 characters, so the cluster Service name `hivemq-%s-cluster` of the initial member list would exceed the DNS label limit of 63 characters" .Release.Name .Release.Name) -}}
+    {{- end -}}
+    {{- $deployedConfig := include "hivemq-platform.deployed-hivemq-configuration" . -}}
+    {{- if contains "<initial-members>" $deployedConfig -}}
+        {{- $deployedMemberCount := len (regexFindAll "<member>" $deployedConfig -1) -}}
+        {{- if ne $deployedMemberCount (int $clustering.initialMemberCount) -}}
+            {{- fail (printf "\nThe `dataIntelligence.clustering.initialMemberCount` value (%d) cannot be changed on an existing installation (deployed with %d initial members)" (int $clustering.initialMemberCount) $deployedMemberCount) -}}
+        {{- end -}}
+    {{- end -}}
+{{- else if and .Values.config.create (not .Values.config.overrideHiveMQConfig) -}}
```
Donnerbart, line 246: This branch is taken by every default installation: `config.create` defaults to `true` and `config.overrideHiveMQConfig` is empty, so `deployed-hivemq-configuration` runs even when no `dataIntelligence` value is set at all. That helper issues both a `ConfigMap` and a `Secret` lookup, and `validate-persistence` adds a third lookup for the `HiveMQPlatform` resource regardless of whether `persistence` is configured.

Until now the only `lookup` in this chart was the CSI driver read, which is guarded by both a `csi` volume type and `Capabilities.APIVersions.Has "hivemq.com/v1"`. Helm returns an empty value for `NotFound` but propagates every other API error, so an installing identity without `get` on Secrets or on `hivemqplatforms` in the release namespace now fails to render a chart that rendered fine before, while using none of the new values.

Either restrict these lookups to the value combinations that need them, or document the added permissions as a requirement of this release.

charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -94,24 +94,287 @@ Usage: {{ include "hivemq-platform.has-license" . }}
...
+    {{- end -}}
+    {{- if not $clustering.initialMemberCount -}}
+        {{- fail (printf "\nThe `dataIntelligence.clustering.initialMemberCount` value is required when HiveMQ Data Intelligence clustering is enabled") -}}
+    {{- end -}}
+    {{- if and (gt (int .Values.nodes.replicaCount) 0) (gt (int $clustering.initialMemberCount) (int .Values.nodes.replicaCount)) -}}
+        {{- fail (printf "\nThe `dataIntelligence.clustering.initialMemberCount` value (%d) cannot be greater than the `nodes.replicaCount` value (%d). Increase `nodes.replicaCount`, or lower `initialMemberCount` before the initial installation, as it cannot be changed afterwards" (int $clustering.initialMemberCount) (int .Values.nodes.replicaCount)) -}}
+    {{- end -}}
+    {{- if and (eq (int $clustering.initialMemberCount) 1) (gt (int .Values.nodes.replicaCount) 1) -}}
+        {{- fail (printf "\nThe `dataIntelligence.clustering.initialMemberCount` value must be at least 2 when the `nodes.replicaCount` value (%d) is greater than 1. Set `initialMemberCount` to at least 2, or run a single replica" (int .Values.nodes.replicaCount)) -}}
```
Donnerbart: `initialMemberCount: 1` combined with `nodes.replicaCount: 1` passes both guards and installs with a single initial member. Scaling that installation up afterwards has no supported path:

* `nodes.replicaCount: 3` with `initialMemberCount: 1` fails here, advising an `initialMemberCount` of at least 2.
* Raising `initialMemberCount` to 2 then fails the immutability check below with "cannot be changed on an existing installation (deployed with 1 initial members)".

The installation stays fixed at one replica until it is reinstalled, and neither message says so. Either reject `initialMemberCount: 1` when clustering is enabled, or extend this message the way the `nodes.replicaCount` message above already does ("before the initial installation, as it cannot be changed afterwards").

charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -94,24 +94,287 @@ Usage: {{ include "hivemq-platform.has-license" . }}
...
+    {{- if not (contains "HIVEMQ_DATA_INTELLIGENCE_CONNECTION_STRING" .Values.config.overrideStatefulSet) -}}
+        {{- fail (printf "\nThe `dataIntelligence.server.connectionStringSecretName` value requires the `HIVEMQ_DATA_INTELLIGENCE_CONNECTION_STRING` environment variable, which the `config.overrideStatefulSet` value does not define. Define the environment variable in your custom StatefulSet, or remove one of the two values") -}}
+    {{- end -}}
+{{- end -}}
+{{- if and $hasConnectionString .Values.config.overrideHiveMQConfig -}}
+    {{- if not (contains "HIVEMQ_DATA_INTELLIGENCE_CONNECTION_STRING" .Values.config.overrideHiveMQConfig) -}}
+        {{- fail (printf "\nThe `dataIntelligence.server.connectionStringSecretName` value requires the `HIVEMQ_DATA_INTELLIGENCE_CONNECTION_STRING` environment variable, which the `config.overrideHiveMQConfig` value does not reference. Reference it as `${HIVEMQ_DATA_INTELLIGENCE_CONNECTION_STRING}` in your custom HiveMQ configuration, or remove one of the two values") -}}
+    {{- end -}}
+{{- end -}}
+{{- if and $hasConnectionString (not .Values.config.create) (not .Values.config.overrideHiveMQConfig) -}}
```
Donnerbart: `templates/hivemq-configuration.yml` renders under `{{- if .Values.config.create }}`, so with `config.create=false` the `config.overrideHiveMQConfig` value is never rendered anywhere.

For `config.create=false` plus `config.overrideHiveMQConfig` plus `dataIntelligence.server.connectionStringSecretName`, the check above validates the value that is ignored, and this check is skipped for the configuration the broker loads. When that existing ConfigMap or Secret carries no `${HIVEMQ_DATA_INTELLIGENCE_CONNECTION_STRING}` reference, the connection string is dropped without a warning, which is the case this validation exists to catch.

```suggestion
{{- if and $hasConnectionString (not .Values.config.create) -}}
```

charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -94,24 +94,287 @@ Usage: {{ include "hivemq-platform.has-license" . }}
...
+kind configured by the `config.createAs` value and falling back to the other kind, so a switched
+`config.createAs` value is still found and a stale leftover of the other kind cannot shadow the
+current configuration. Returns an empty string when no configuration is deployed or when rendering
+client-side (`helm template`), where `lookup` returns nothing.
+Usage: {{ include "hivemq-platform.deployed-hivemq-configuration" . }}
+*/}}
+{{- define "hivemq-platform.deployed-hivemq-configuration" -}}
+{{- $configName := include "hivemq-platform.configuration-name" . -}}
+{{- $configMap := lookup "v1" "ConfigMap" .Release.Namespace $configName -}}
+{{- $secret := lookup "v1" "Secret" .Release.Namespace $configName -}}
```
Donnerbart: Both kinds are read on every render, before the `config.createAs` branch below decides which one is used. The second read only matters as a fallback when the preferred kind is missing, so issuing it inside the `else if` branches would halve the API calls and keep the Secret read off the ConfigMap default path.

charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -94,24 +94,287 @@ Usage: {{ include "hivemq-platform.has-license" . }}
...
+        {{- if and $service.exposed (eq (int64 $service.containerPort) $clusteringPort) -}}
+            {{- fail (printf "\nThe `dataIntelligence.clustering.port` value (%d) already exists as the container port of the exposed service type `%s`" $clusteringPort $service.type) -}}
+        {{- end -}}
+    {{- end -}}
+    {{- if gt (len .Release.Name) 48 -}}
+        {{- fail (printf "\nThe release name `%s` exceeds 48 characters, so the cluster Service name `hivemq-%s-cluster` of the initial member list would exceed the DNS label limit of 63 characters" .Release.Name .Release.Name) -}}
+    {{- end -}}
+    {{- $deployedConfig := include "hivemq-platform.deployed-hivemq-configuration" . -}}
+    {{- if contains "<initial-members>" $deployedConfig -}}
+        {{- $deployedMemberCount := len (regexFindAll "<member>" $deployedConfig -1) -}}
```
Donnerbart: This counts every `<member>` in the whole deployed configuration, not only the ones inside `<initial-members>`. It is correct as long as the initial member list stays the only place that renders a `<member>` element, and it silently miscounts as soon as another section renders one. Scoping the match to the block removes that dependency:

```suggestion
        {{- $initialMembers := regexFind "(?s)<initial-members>.*?</initial-members>" $deployedConfig -}}
        {{- $deployedMemberCount := len (regexFindAll "<member>" $initialMembers -1) -}}
```

charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -94,24 +94,287 @@ Usage: {{ include "hivemq-platform.has-license" . }}
...
+current configuration. Returns an empty string when no configuration is deployed or when rendering
+client-side (`helm template`), where `lookup` returns nothing.
+Usage: {{ include "hivemq-platform.deployed-hivemq-configuration" . }}
+*/}}
+{{- define "hivemq-platform.deployed-hivemq-configuration" -}}
+{{- $configName := include "hivemq-platform.configuration-name" . -}}
+{{- $configMap := lookup "v1" "ConfigMap" .Release.Namespace $configName -}}
+{{- $secret := lookup "v1" "Secret" .Release.Namespace $configName -}}
+{{- if eq .Values.config.createAs "Secret" -}}
+    {{- if $secret -}}
```
Donnerbart: The fallback covers a missing resource but not a missing key. When the preferred kind exists without a `config.xml` entry, `dig` returns an empty string and the other kind is never consulted, so the caller sees an empty configuration and every check built on it silently passes. Testing the extracted value rather than the resource would cover both cases, and combines with the lazy lookup noted on the two `lookup` lines above.

Donnerbart (commented):
A few notes on the new validation helpers and the lookups they introduce.

charts/hivemq-platform/templates/_helpers.tpl
```diff
@@ -94,24 +94,287 @@ Usage: {{ include "hivemq-platform.has-license" . }}
...
+   `helm template`, where `lookup` returns nothing.
+Usage: {{ include "hivemq-platform.validate-data-intelligence" . }}
+*/}}
+{{- define "hivemq-platform.validate-data-intelligence" -}}
+{{- if or (.Values.pulse).create (.Values.pulse).name (.Values.pulse).data (.Values.pulse).overridePulseConfig -}}
+    {{- fail (printf "\nThe `pulse` values are retired: the HiveMQ Data Intelligence configuration moved into the HiveMQ configuration (config.xml). Use the `dataIntelligence` values instead and remove the `pulse` values (`--set pulse=null` when reusing values)") -}}
+{{- end -}}
+{{- $hasConnectionString := include "hivemq-platform.has-data-intelligence-connection-string" . -}}
+{{- if and $hasConnectionString .Values.config.overrideStatefulSet -}}
+    {{- if not (contains "HIVEMQ_DATA_INTELLIGENCE_CONNECTION_STRING" .Values.config.overrideStatefulSet) -}}
```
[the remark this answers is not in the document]
afalhambra-hivemq: Improved this already. Not much we can do here

afalhambra-hivemq (approved):

PR #1077, opened 2026-09-17
Authenticate the kubeconform plugin install against the GitHub API

Donnerbart:
## Summary

- Pass `GITHUB_TOKEN` to the kubeconform plugin install in the edge, platform and swarm verify workflows. The plugin's install hook looks up the latest kubeconform release via the GitHub API and fails with `Failed to get DOWNLOAD_URL` when the unauthenticated request is rate limited.


afalhambra-hivemq (approved):
Good one! It fails from time to time, and probably it was due to this sometimes

